PR #12, opened 2019-09-16
Update GoDoc for Azure.

negz:

This documentation is used to generate our user facing documentation, both in CRDs and on the website. I've edited it for tone, and to generally make sense where it did not.

### Checklist

I have:
- [ ] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.

### DIFF
--- a/azure/apis/cache/v1alpha2/redis_types.go
+++ b/azure/apis/cache/v1alpha2/redis_types.go
@@ -58,7 +58,9 @@ const (
 	SupportedRedisVersion = "3.2"
 )
 
-// RedisParameters defines the desired state of Redis
+// RedisParameters define the desired state of an Azure Redis cluster. Most
+// fields map directly to an Azure Redis resource:
+// https://docs.microsoft.com/en-us/rest/api/redis/redis/create#redisresource
 type RedisParameters struct {
 	// ResourceGroupName in which to create this resource.
 	ResourceGroupName string `json:"resourceGroupName"`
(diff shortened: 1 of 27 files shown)


azure/apis/cache/v1alpha2/redis_types.go
```diff
@@ -122,15 +122,21 @@ type SKUSpec struct {
...
+	runtimev1alpha1.ResourceSpec `json:",inline"`
+	RedisParameters              `json:",inline"`
+}
+
+// A RedisStatus represents the observed state of a Redis.
 type RedisStatus struct {
 	runtimev1alpha1.ResourceStatus `json:",inline"`
 
-	State   string `json:"state,omitempty"`
-	Message string `json:"message,omitempty"`
```
hasheddan: Is the `Message` field supposed to be completely removed?

azure/apis/compute/v1alpha2/types.go
```diff
@@ -25,84 +25,104 @@ import (
...
 )
 
-// AKSClusterParameters define the configuration for AKS cluster resources
+// AKSClusterParameters define the desired state of an Azure Kubernetes Engine
+// cluster.
 type AKSClusterParameters struct {
-	// ResourceGroupName is the name of the resource group that the cluster will be created in
-	ResourceGroupName string `json:"resourceGroupName"` //--resource-group
+	// ResourceGroupName is the name of the resource group that the cluster will
+	// be created in
```
hasheddan: I notice a number of these comments do not end with a period. Is that intentional?

azure/apis/network/v1alpha2/types.go
```diff
@@ -1,17 +1,16 @@
 /*
 Copyright 2019 The Crossplane Authors.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of the
+License at
```
hasheddan: Looks like license header was changed here but not elsewhere, is that intended?

azure/apis/storage/v1alpha2/account.go
```diff
@@ -379,7 +384,7 @@ func toStorageSkuCapability(s skuCapability) storage.SKUCapability {
 	}
 }
 
-// Sku the Sku of the storage account.
+// Sku of an Azure Blob Storage Account.
```
hasheddan: I apologize on behalf of Azure lol

azure/apis/network/v1alpha2/types.go
negz: Ah I think this was me trying out that reformat extension you recommended. Will fix.

azure/apis/compute/v1alpha2/types.go
negz: I'd prefer them to end with periods, but I'm sure I missed them in a bunch of places since I'm trying to power through all this pretty quickly.

hasheddan: Sounds good! I'll try to go through and catch some, probably after v0.3 :)

azure/apis/cache/v1alpha2/redis_types.go
negz: Yeah that was intentional - it's never actually set anywhere. I suspect this is the case for a lot of our resources, but this one I checked.

negz:
@hasheddan PTAL - I believe I addressed all the comments except for the missing periods, which I think we can fix up as we spot them (or during code generation).

Speaking of code generation, one interesting observation is that many of the Azure SDK types seem to be a little lacking in the comment department, with comments like `// Foo - Gets or sets the Foo`, or just `// Foo..`, so it may not be a given that we can get high quality documentation of our fields simply by inspecting the SDK types. :(

PR #19, opened 2019-09-20
change binary name and remove unnecessary directories

### LINKED ISSUE #6: Stack binary should be named "stack-azure"
### What happened?

The entrypoint binary for `stack-azure` appears to be named `crossplane`. This breaks at least the `make run` command, which expects the output binary to be the same as the project name.

### How can we reproduce it?

`make run`

Crossplane version:

hasheddan:

Renames binary name from `crossplane` to `stack-azure`. Will open follow-up PR's for `stack-aws` and `stack-gcp` when merged.

Fixes #6 

### Checklist

I have:
- [x] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.


soorena776:
@hasheddan I was thinking about the same thing in stack-aws
Shouldn't we change the `cmd/crossplane` simply to `cmd/stack`?

soorena776:
@hasheddan also, fyi I am working on this
https://github.com/crossplaneio/stack-aws/issues/9

I could later apply the same change to other stack repositories too

hasheddan:
@soorena776 changing it to `stack` may be a good idea for consistency across the stacks and being able to reuse dockerfiles and makefiles! @negz does this sound right to you? Noticed you suggested `stack-azure` in #6

negz:
@soorena776 @hasheddan I don't feel strongly about whether we name the binaries `stack` or `stack-${cloud}`. I suggested the latter because I thought some of our build tooling assumed the project name must match the binary name _and_ repo name, but I could be wrong.

hasheddan:
@negz I guess the fact that it was working when we created the binary named `crossplane` suggests that they can differ

negz:
>  the fact that it was working when we created the binary named crossplane suggests that they can differ

@hasheddan I recall that parts of the build tooling, (at least `make run`) are currently broken by the binary being named `crossplane`, because they assume it will be named `stack-azure`. It's likely this is more of an issue with the build tooling than anything else.

negz:
Adding the Stacks folks to this PR to get their take on what the binary should be named. If it doesn't affect anything Stack-ey or build-tool-ey then this LGTM.

suskin:
From a build tooling perspective, the current tooling does make some assumptions about the build process (it assumes modules are used), and it also assumes that the binary is named `manager`. But, we have been assuming that these assumptions will be too inflexible in the long run. I expect this will become something for which the build tool has no opinion, but for which we have a recommended best practice.

From the perspective of the stack manager, there are no opinions on what the entrypoint binary should be named; it only expects that there is one.

If I were choosing a name from scratch, I would probably try to avoid using the word "stack" in the name, because Stack is related to the packaging format and delivery channel. I think of the binary (and its name) as related to running the controller, and therefore not related to any particular packaging format and delivery channel. I would use something that seems right regardless of how the controller happens to be packaged and delivered. I'm brainstorming now, but options for names could include "controller", "manager" (I am pulling this from kubebuilder), "azure", "crossplane-azure", "provider", "crossplane-provider", "cloud-provider", and so on.

negz:
I like the `crossplane-azure` suggestion, so we'd have `crossplane-azure`, `crossplane-gcp`, etc. I'm not bullish on this, but it would be good to pick a name and move on with our lives. :)

jbw976:
the `crossplane-azure`, `crosspalne-gcp`, `crossplane-aws` suggestion sounds great to me.

soorena776:
I just talked to @suskin, and I think giving more specific name to the binary, like `crossplane-azure-stack` or `crossplane-azure` makes sense. However for consistency across different stack repositories we think the folder name should be consistent as `stack`, and this renaming should be done in the Dockerfile, like:

```Dockerfile
ADD stack /usr/local/bin/crossplane-azure-stack
...
ENTRIPOINT ["crossplane-azure-stack"]
```

hasheddan:
cc @negz @soorena776 

updated to match work done in crossplaneio/stack-aws#9

negz:
Happy to merge if @soorena776 can confirm this matches his changesets by approving.

PR #29, opened 2019-10-04
Implement Integration Tests

hasheddan:

Fixes crossplaneio/crossplane#759

This initial implementation just installs the stacks and makes sure its pods are running. In the future, we may want to actually create some custom resources.

### TODO

- [x] Handle case where stack is installed before `kube-apiserver-[kind_cluster_name]-control-plane` / `kube-controller-manager-[kind_cluster_name]-control-plane` / `etcd-[kind_cluster_name]-control-plane` / `kube-scheduler-[kind_cluster_name]-control-plane`
- [x] Add timeouts to manual waits

### Checklist

I have:
- [x] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.

### DIFF
--- a/cluster/local/integration_tests.sh
+++ b/cluster/local/integration_tests.sh
@@ -32,20 +32,68 @@ echo_error(){
     exit 1
 }
 
+# k8s watchers
+wait_for_deployment_create(){
+    local timeout=$1
+    local counter=0
+    echo -n "waiting for deployment in $2..." >&2
+    while "${KUBECTL}" -n $2 get deployments -o yaml | grep -q  'items: \[\]'; do
+        if [ "$counter" -ge "$timeout" ]; then echo "TIMEOUT"; exit -1; else (( counter+=5 )); fi
+        echo -n "." >&2
+        sleep 5
+    done
+}
+
+wait_for_pods_in_namespace(){
+    local timeout=$1
+    shift
+    namespace=$1
+    shift 
+    arr=("$@")
+    local counter=0
+    for i in "${arr[@]}";
+        do
+            echo -n "waiting for pod $i in namespace $namespace..." >&2
+            while ! ("${KUBECTL}" -n $namespace get pod $i) &>/dev/null; do
+                if [ "$counter" -ge "$timeout" ]; then echo "TIMEOUT"; exit -1; else (( counter+=5 )); fi
+                echo -n "." >&2
+                sleep 5
+            done
+            echo "FOUND POD!" >&2
+        done
+}
+
 # ------------------------------
 projectdir="$( cd "$( dirname "${BASH_SOURCE[0]}")"/../.. && pwd )"
 
 # get the build environment variables from the special build.vars target in the main makefile
 eval $(make --no-print-directory -C ${projectdir} build.vars)
 
+# ------------------------------
+
+HELM_VERSION="helm-v2.10.0"
+HOST_PLATFORM="${HOSTOS}_${HOSTARCH}"
+TOOLS_DIR="${CACHE_DIR}/tools"
+TOOLS_HOST_DIR="${TOOLS_DIR}/${HOST_PLATFORM}"
+HELM="${TOOLS_HOST_DIR}/${HELM_VERSION}"
+echo_step "installing helm ${HOSTOS}-${HOSTARCH}"
+mkdir -p ${TOOLS_HOST_DIR}/tmp-helm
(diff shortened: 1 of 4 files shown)


cluster/local/integration_tests.sh
```diff
@@ -0,0 +1,304 @@
+#!/usr/bin/env bash
+set -e
```
soorena776: Do you think we should refactor this, and move the common part to build repository? It feels like only line 260 onward should be in this repo

displague: I was wondering the same.  It may be easier to bring this benefit to the other crossplaneio managed stacks if it were in the build project.

As a standalone test, this may be easier to inform stack developers on how to create integration tests of their own, however the stack-azure build process is already dependent on the crossplaneio/build project.  Because of the dependency, I don't think this potential value stands.  I agree that this would be best homed in crossplaneio/build.

I would be fine with a merge into this repo if we wanted to follow up with the relocation later.

hasheddan: It seems like line 207 onward may be necessary to keep independent. However, I am not sure any of it is generic enough to be in the build submodule. It might be useful to get this implemented and then investigate how we can standardize on it across stacks, I definitely have some ideas. Update the build submodule can result in long cycles as I know you are aware of :)

cluster/local/integration_tests.sh
```diff
@@ -0,0 +1,278 @@
...
+
+# setting up colors
+BLU='\033[0;34m'
+YLW='\033[0;33m'
+GRN='\033[0;32m'
+RED='\033[0;31m'
+NOC='\033[0m' # No Color
+echo_info(){
```
suskin: Nice! I guess I have the same question as other people, which is how we can make it reusable across repos.

cluster/local/integration_tests.sh
```diff
@@ -0,0 +1,278 @@
...
+"${KUBECTL}" -n kube-system rollout status deploy/tiller-deploy --timeout=2m
+
+# wait for kind pods
+echo_step "wait for kind pods"
+kindpods=("kube-apiserver-${BUILD_REGISTRY}-inttests-control-plane" "kube-controller-manager-${BUILD_REGISTRY}-inttests-control-plane" "kube-scheduler-${BUILD_REGISTRY}-inttests-control-plane")
+wait_for_pods_in_namespace 120 "kube-system" "${kindpods[@]}"
+
+# install crossplane from alpha channel
+echo_step "installing crossplane from alpha channel"
+"${HELM}" repo add crossplane-alpha https://charts.crossplane.io/alpha
```
suskin: If we are testing from the alpha channel, how do we test a version which we are working on locally? Or a version on a feature branch which is in review?

cluster/local/integration_tests.sh
hasheddan: IMO we should eventually develop some sort of integration testing that is external to these stack repos. As long as we are doing it with scripts and Jenkinsfiles in each of the repositories, then there is potential for testing across stacks to be inconsistent. Moreover, as more and more stacks are created, it would be nice if they could just have a `Crosstest` configuration file or something of that nature (which could potentially be generated by something like the crossplane-cli) that allows them to plugin to a test suite in a simple and straightforward fashion. I have lots of ideas around this that would probably be good to reflect in an issue in `crossplaneio/crossplane`. However, in the meantime I would like to have even this minimal integration testing in place for the three existing stacks because I don't like the idea that PR's are merged without being tested.

cluster/local/integration_tests.sh
hasheddan: Right now the integration tests only run against alpha, but I think enabling other revisions would be an enhancement. These integration tests are written to do the bare minimum of making sure that install and uninstall are successful. If there is desire to test against other local crossplane branches then this portion could be replaced for local e2e. As with the comment above, I think that there is a need to develop a much more robust testing engine in the future for stacks.

negz:
I trust @soorena776's LGTM on this PR, given he has much more context on our existing integration test patterns and challenges than I do. @hasheddan are you ready for this to be merged?

One observation I have about our existing corpus of integration tests is that they rely heavily on Helm and shell scripts. I like to imagine integration tests as existing on a spectrum from treating the code under test as a binary black box as these tests do, through to being a harness that acts as an alternative to `main.go`, loading an exercising only the parts of the codebase it wishes to test, similar to the style of test kubebuilder encourages.

Either way, my suspicion is that we may end up with a more sustainable, maintainable integration test framework if we were to approach this by using Go tests (i.e. the `testing` package) to write our integration tests. This could enable @hasheddan's idea of a "Crossplane integration testing framework" in the form of a Go library.

hasheddan:
@negz ready and agree with your sentiment here!

PR #35, opened 2019-10-21
Migrate Azure to Simple Resource Class Selection

### LINKED ISSUE #34: Implement Simple Resource Class Selection for Azure

Dynamic provisioning is complicated. See https://github.com/crossplaneio/crossplane/pull/926 for full context.

Implement the patterns described in https://github.com/crossplaneio/crossplane/pull/926 for the Azure Stack. Specifically:

* Make providers, classes, and managed resources cluster scoped.
* Match classes to claims using label selectors
* Fall back to using a resource class annotated as the default.

This will depend on https://github.com/crossplaneio/crossplane/pull/927 and https://github.com/crossplaneio/crossplane-runtime/pull/48.

negz:

Fixes #34.

Apologies in advance reviewers for this huge and unwieldy PR. This PR:

* Adds comment markers to make all providers, resources, and classes cluster scoped.
* Regenerates all generated code and manifests (using https://github.com/crossplaneio/crossplane-tools/pull/7) to support cluster scoped managed resources.
* Updates all controller logic to reflect that providers and managed resources are no longer namespaced, and thus they must specify which namespace their secrets live in.
* I tried to keep these changes minimal but after struggling for many hours with the old SQL server code and tests I decided it would be faster to port them to the managed resource reconciler since @muvaf is hoping to bring this to v1beta1 this week.

So far I have tested:
* Creating a `Redis` by submitting a `RedisCluster` that references a `RedisClass`.
* Creating a `Redis` by submitting an unopinionated `RedisCluster` claim that falls back to using the `RedisClass` annotated as default.
* Creating several `RedisClass` classes with the same label and ensuring `RedisCluster` claims that select those labels are allocated a random matching class.

I've also tested: 
* Creating (and connecting to) a PostgreSQL database.
* Creating a MySQL database.
* Creating a storage account and container.
* Creating an AKS cluster.

I don't intend to manually test claim scheduling and defaulting logic for these kinds, because it's all the same reconciler under the hood. At this stage I'm leaning toward not explicitly testing all the connectivity primitives. The changes to their types were very mechanical - they have no claims or connection secrets so it was just updating the logic that gets the provider secret. We should still test them holistically once all our in flight changes land - I imagine any bugs I might have introduced here would be quickly fixable.

### Checklist

I have:


negz:
@hasheddan PTAL. Mostly looking for your input on this one, but I've also copied @jbw976 for his historical familiarity with the AKS and SQL Server controllers.

negz:
Also CCing @muvaf on this one, since I jumped the gun on porting the Azure database controller a little.

pkg/controller/compute/claim.go
```diff
@@ -103,8 +151,15 @@ func ConfigureAKSCluster(_ context.Context, cm resource.Claim, cs resource.NonPo
 	if spec.NodeCount == nil {
 		spec.NodeCount = to.IntPtr(v1alpha2.DefaultNodeCount)
 	}
-	spec.WriteServicePrincipalSecretTo = corev1.LocalObjectReference{Name: fmt.Sprintf("principal-%s", cm.GetUID())}
-	spec.WriteConnectionSecretToReference = corev1.LocalObjectReference{Name: string(cm.GetUID())}
+
+	spec.WriteServicePrincipalSecretTo = runtimev1alpha1.SecretReference{
```
hasheddan: Do we feel good as a universal best practice to write *all* secrets produced by managed resources to the same namespace? I don't see any problems with it, I just wanted to bring up in case you could think of a situation where we would want different behavior.

pkg/clients/sql.go
```diff
@@ -329,189 +210,95 @@ func MySQLVirtualNetworkRuleStatusFromAzure(az mysql.VirtualNetworkRule) azuredb
...
 // NewPostgreSQLServerClient creates and initializes a PostgreSQLServerClient instance.
-func NewPostgreSQLServerClient(provider *v1alpha2.Provider, clientset kubernetes.Interface) (*PostgreSQLServerClient, error) {
-	client, err := NewClient(provider, clientset)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create Azure client: %+v", err)
-	}
-
-	postgreSQLServerClient := postgresql.NewServersClient(client.SubscriptionID)
-	postgreSQLServerClient.Authorizer = client.Authorizer
+func NewPostgreSQLServerClient(c *Client) (*PostgreSQLServerClient, error) {
```
hasheddan: This applies to a number of the changes to `clients` where they are being refactored to have the `azure.Client` type passed in. I generally prefer passing in something like `[]byte` and then calling to the general `NewClient()` func within `NewPostgreSQLServerClient`. It reduces the code in the managed controllers (slightly, I'll admit), but generally feels cleaner. There is also the option of just doing all of the parsing each time in the actual `New___()` function as is currently being done in the `Redis` client (https://github.com/negz/stack-azure/blob/3b6e68a65f0d1507111937eff1be00a6c899b40b/pkg/clients/redis/redis.go#L45). I may be missing a clear reason why you want to pass in `Client` instead here, so please feel free to push back on my thoughts here :) One reason it may be useful is with something like the AKS cluster where multiple clients are being used. In general, I would prefer consistency with the `New___()` funcs though.

pkg/controller/database/mysqlserver/managed.go
```diff
@@ -0,0 +1,212 @@
...
+	s, ok := mg.(*v1alpha2.MysqlServer)
+	if !ok {
+		return resource.ExternalObservation{}, errors.New(errNotMysqlServer)
+	}
+
+	external, err := e.client.GetServer(ctx, s)
+	if azure.IsNotFound(err) {
+		// Azure SQL servers don't exist according to the Azure API until their
+		// create operation has completed, and Azure will happily let you submit
+		// several subsequent create operations for the same server. Our create
```
hasheddan: Wow, I did not know this. Thank you for identifying this! I do not love this kind of API behavior by the Azure SDK :/

pkg/controller/database/postgresqlserver/managed.go
```diff
@@ -0,0 +1,212 @@
...
+		For(&v1alpha2.PostgresqlServer{}).
+		Complete(r)
+}
+
+func newClient(credentials []byte) (azure.PostgreSQLServerAPI, error) {
+	ac, err := azure.NewClient(credentials)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot create Azure client")
+	}
+	pc, err := azure.NewPostgreSQLServerClient(ac)
```
hasheddan: In regards to my earlier comment, maybe it is more informative to have these two separate client creation steps? I am not sure.. I don't know if there would ever be a case in which we could create the general client but then fail to create the service specific one.

pkg/controller/storage/account/claim.go
```diff
@@ -35,11 +34,65 @@ import (
...
+// A ClaimDefaultingController reconciles Bucket claims that omit their resource
+// ref, class ref, and class selector by choosing a default Azure Account
+// resource class if one exists.
+type ClaimDefaultingController struct{}
+
+// SetupWithManager sets up the ClaimDefaultingController using the supplied
+// manager.
+func (c *ClaimDefaultingController) SetupWithManager(mgr ctrl.Manager) error {
+	name := strings.ToLower(fmt.Sprintf("defaulter.%s.%s.%s",
+		storagev1alpha1.BucketKind,
```
hasheddan: A more general thought that should not be addressed in this PR: I am not sure we should continue to support dynamic provisioning of a storage account with a bucket claim. It feels like we should either just make the account have to be statically provisioned, or develop some set of claims for organizational things like the storage accounts / resource groups / etc... not sure we could ever be that general with them though.

apis/database/v1alpha2/sql_types.go
```diff
@@ -212,13 +174,6 @@ type SQLServerStatus struct {
 
 	// Endpoint of the MySQL Server instance used in connection strings.
 	Endpoint string `json:"endpoint,omitempty"`
-
-	// RunningOperation stores any current long running operation for this
-	// instance across reconciliation attempts.
-	RunningOperation string `json:"runningOperation,omitempty"`
-
-	// RunningOperationType is the type of the currently running operation.
-	RunningOperationType string `json:"runningOperationType,omitempty"`
```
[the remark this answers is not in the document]
negz: Right, these are not exposed by the Azure API (as fields of the SQL server object). We used them to persist the running operation that was returned when we asked to create or delete a server. I'm guessing because it was the only obvious way to know that we'd asked for something to happen to the server given that AFAICT changes aren't reflected immediately in the API _and_ there's no way to query for running operations. (CC @jbw976 for history.)

We prefer to act based on real world observations rather than recorded status, so I switched to checking whether the name is taken to determine whether an SQL server is currently being created, which is the only operation we need to be aware of with the current implementation.

I agree that there will be more changes necessary to get this to v1beta1. My goal here was to unblock myself on the simple resource class selection front so I wanted to make minimal changes to the API and functionality. I believe @muvaf is going to be working on getting this to v1beta1 in the coming weeks, so hopefully this change moves the controller slightly in that direction.

pkg/clients/sql.go
negz: I think I agree that it's better to just pass the credential bytes in same as we do elsewhere. Will update.

pkg/controller/database/mysqlserver/managed.go
negz: Yeah it's not great. Without this check we'd repeatedly try to create the server. The first create would win, and the other 3-4 would fail, but we'd generate a new random password each time and end up persisting a password that did not actually work.

pkg/controller/database/postgresqlserver/managed.go
negz: Yeah, mostly the service specific client creation ones are doing very little. I think I'll change this to match your preference.

pkg/controller/storage/account/claim.go
negz: Yeah I was thinking the same thing while updating this controller. Would you mind raising an issue to track?

negz:
@hasheddan I notice the integration tests are consistently failing for this PR, but couldn't immediately determine why. Can you see a reason why this PR would break the tests?

hasheddan:
It appears as though the deployment of the stack is failing. I can checkout this branch and test locally.

pkg/clients/sql.go
negz: Ergh, actually I just gave this a shot and found it harder to update the AKS controller than I expected due to how its tests work. I still like this direction but would prefer to raise an issue for it rather than try include it in this PR.

negz:
@hasheddan This is the issue I encountered when trying to test out AKS:

```
Message:               failed to start create operation for AKS cluster default-app-kubernetes-gnxv5: containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="LinkedInvalidPropertyId" Message="Property id '' at path 'properties.agentPoolProfiles[0].vnetSubnetID' is invalid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'."
```

Using this class:
```yaml
---
apiVersion: compute.azure.crossplane.io/v1alpha2
kind: AKSClusterClass
metadata:
  name: akscluster-standard
specTemplate:
  writeConnectionSecretsToNamespace: crossplane-system
  # NOTE(negz): This field being required by resource classes is a bug. It is
  # ignored when dynamically provisioning AKS clusters.
  writeServicePrincipalTo:
    namespace: unused
    name: unused
  resourceGroupName: group-westus-1
  location: West US
  version: "1.12.8"
  nodeCount: 1
  nodeVMSize: Standard_B2s
  dnsNamePrefix: crossplane-aks
  disableRBAC: false
  providerRef:
    name: example
  reclaimPolicy: Delete
```

Which is https://github.com/crossplaneio/crossplane/blob/12be413b9/cluster/examples/compute/kubernetescluster/azure/resource-class.yaml updated to support simple resource class selection.

hasheddan:
AKS cluster fixed with #37 

@negz I investigated the integration tests and the issue is that it uses the alpha version of Crossplane (v0.3.0). This version does not like that `stack-azure` is attempting to install cluster-scoped resources. The exact error is:
```
crossplane: error: failed to unpack stacks: failed to walk Stack filesystem: Stack CRD "/ext-pkg/.registry/resources/azure.crossplane.io_providers.crd.yaml" must be namespaced scope
```

This will also presumably fail for the updates to the other stacks as well. This will need to be fixed in Crossplane (if it isn't already), then we will need to address in integration tests. Do you think it would also be wise to always use the latest master version of Crossplane instead for now? In the long-run I would like to be able to test against Crossplane versions based on the most recent declared compatible version of each stack. In the mean time, we could just run integration tests against alpha and master and make judgement calls about which of them needs to be able to pass for merge. I have bandwidth to implement this week.

pkg/controller/compute/claim.go
negz: I'm not sure. I think this is the only managed resource that creates more than one secret, and it's not super obvious to me why it writes the service principal to a distinct secret from the connection secret. I suppose probably because it has to read it back? I could see the case that we should have a distinct `writeServicePrincipalSecretsToNamespace` class option for this case if we need to keep doing this, but I'd rather tackle that in future.

pkg/controller/database/postgresqlserver/managed.go
negz: I agree with the direction, but ended up punting on this in the interest of time.

apis/storage/v1alpha2/types.go
```diff
@@ -80,8 +81,8 @@ type AccountList struct {
 // An AccountClassSpecTemplate is a template for the spec of a dynamically
 // provisioned Account.
 type AccountClassSpecTemplate struct {
-	runtimev1alpha1.NonPortableClassSpecTemplate `json:",inline"`
-	AccountParameters                            `json:",inline"`
+	runtimev1alpha1.ClassSpecTemplate `json:",inline"`
+	AccountParameters                 `json:",inline"`
```
hasheddan: Assuming `forProvider` will be a follow-up PR?

negz: Yeah, don't want to make that change in this already huge PR.

PR #58, opened 2019-11-05
Redis v1beta1

### LINKED ISSUE #50: Azure Redis v1beta1 Quality
Redis should be refactored to use generic managed reconciler and adopt patterns in https://github.com/crossplaneio/crossplane/pull/840

Part of https://github.com/crossplaneio/crossplane/issues/863

muvaf:

Refactors Redis to adopt patterns introduced in https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md

Part of https://github.com/crossplaneio/crossplane/issues/863
Fixes #50 

### Checklist

I have:


apis/cache/v1alpha3/redis_types.go
```diff
@@ -56,101 +29,138 @@ const (
 	SupportedRedisVersion = "3.2"
 )
 
-// RedisParameters define the desired state of an Azure Redis cluster. Most
-// fields map directly to an Azure Redis resource:
+// An SKU represents the performance and cost oriented properties of a
```
hasheddan:
```suggestion
// A SKU represents the performance and cost oriented properties of a
```

pkg/clients/redis/redis.go
```diff
@@ -19,35 +19,38 @@ package redis
...
+	ProvisioningStateCreating  = string(redis.Creating)
+	ProvisioningStateDeleting  = string(redis.Deleting)
+	ProvisioningStateFailed    = string(redis.Failed)
+	ProvisioningStateSucceeded = string(redis.Succeeded)
+	ProvisioningStateUpdating  = string(redis.Updating)
+)
 
 // A Client handles CRUD operations for Azure Cache resources. This interface is
 // compatible with the upstream Azure redis client.
 type Client redisapi.ClientAPI
```
hasheddan: It looks like this can be removed now that we are using `redisapi.ClientAPI` directly.

pkg/controller/cache/redis_test.go
```diff
@@ -73,7 +57,22 @@ const (
...
 
 var (
-	ctx                = context.Background()
+	enableNonSSLPort = true
+	subnetID         = "coolsubnet"
+	staticIP         = "172.16.0.1"
+	shardCount       = 3
+	location         = "coolplace"
+	minTLSVersion    = "1.1"
+	tenantSettings   = map[string]string{"tenant1": "is-crazy"}
```
hasheddan: LOL :)

hasheddan:
Hmm, looks like integration tests are failing on the start-up of the `stack-azure` pod. It is not immediately apparent to me what the issue is, but to debug we could run `make e2e skipcleanup=true` then inspect the logs of the `stack-azure` pod.

muvaf:
> Hmm, looks like integration tests are failing on the start-up of the stack-azure pod

Fixed the issue during my manual test. `SetupWithManager` was registering `RedisClass` instead of `Redis` as managed kind. The next commit will have it.

muvaf:
There is an issue with the updates, similar to what we faced in AWS resources. When you create a Redis instance with no `RedisConfiguration`, it creates and in the next reconcile we acquire the populated defaults. It looks like the following:
```
    redisConfiguration:
      maxclients: "256"
      maxfragmentationmemory-reserved: "12"
      maxmemory-delta: "2"
```

But when you try to `Update`, it returns the following error:
`{"error":{"code":"InvalidRequestBody","message":"The setting 'maxfragmentationmemory-reserved' is not allowed on 'Basic' cache instances.\r\nRequestID=b9babb5f-b84e-4cc0-90a0-eb1e00987695","target":null}}`

It basically says that these are the defaults we define but you cannot change them just because you chose the `Basic` class, which is reasonable except that we don't change it. It's just the same value. To solve this, I'll write a patch function that creates the `UpdateParameters` it wants with only the values that have changed similar to this https://github.com/crossplaneio/stack-aws/blob/master/pkg/clients/rds/rds.go#L139

It seems that only GCP supports `Patch` calls which calculates the diff patch and sends only that to their API, others expect the consumers to come up with a patch object and send it to `Update` endpoint.

apis/cache/v1alpha3/redis_types.go
negz: I feel like "An SKU" is more correct here, for reasons I cannot explain. 🤔

apis/cache/v1alpha3/redis_types.go
```diff
@@ -56,101 +29,138 @@ const (
...
+	Family string `json:"family"`
+
+	// Capacity specifies the size of Redis cache to deploy. Valid values: for C
+	// family (0, 1, 2, 3, 4, 5, 6), for P family (1, 2, 3, 4).
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=6
+	Capacity int `json:"capacity"`
+}
+
+// TODO: trial-error the optional fields.
```
negz: Does this need to be done before we merge this PR?

pkg/clients/redis/redis.go
negz: Is this still used with the change to `NewClient`?

pkg/controller/azure.go
```diff
@@ -46,7 +46,7 @@ func (c *Controllers) SetupWithManager(mgr ctrl.Manager) error {
 		&cache.RedisClaimSchedulingController{},
 		&cache.RedisClaimDefaultingController{},
 		&cache.RedisClaimController{},
-		&cache.RedisController{},
+		&cache.Controller{},
```
negz: I feel like we should standardize on package names like `controller/cache/redis` at some point so we can have `redis.ClaimSchedulingController{}`, `redis.Controller{}`, etc. Not important for the moment.

pkg/controller/cache/redis.go
```diff
@@ -18,297 +18,159 @@ package cache
...
 
 	runtimev1alpha1 "github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1"
-	"github.com/crossplaneio/crossplane-runtime/pkg/logging"
 	"github.com/crossplaneio/crossplane-runtime/pkg/meta"
 	"github.com/crossplaneio/crossplane-runtime/pkg/resource"
 
 	"github.com/crossplaneio/stack-azure/apis/cache/v1alpha3"
 	azurev1alpha3 "github.com/crossplaneio/stack-azure/apis/v1alpha3"
-	azure "github.com/crossplaneio/stack-azure/pkg/clients"
+	common "github.com/crossplaneio/stack-azure/pkg/clients"
```
negz: What is the thinking behind this import alias? I don't recall any other v1beta1 controllers using this naming convention for their clients package.

pkg/controller/cache/redis.go
```diff
@@ -18,297 +18,159 @@ package cache
...
-		r.Status.SetConditions(runtimev1alpha1.ReconcileError(err))
-		return true
-	}
+	name := strings.ToLower(fmt.Sprintf("%s.%s", v1alpha3.RedisKind, v1alpha3.Group))
 
-	r.Status.SetConditions(runtimev1alpha1.ReconcileSuccess())
-	return false
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha3.Redis{}).
```
negz: Will we bump to v1beta1  in a future commit?

apis/cache/v1alpha3/redis_types.go
```diff
@@ -56,101 +29,138 @@ const (
...
+
+// RedisParameters define the desired state of an Azure Redis cluster.
 // https://docs.microsoft.com/en-us/rest/api/redis/redis/create#redisresource
 type RedisParameters struct {
+	// TODO(muvaf): not included in received object.
 	// ResourceGroupName in which to create this resource.
+	// +immutable
 	ResourceGroupName string `json:"resourceGroupName"`
```
negz: Should we introduce a `ResourceGroupNameRef` variant of this field that is an `AttributeReferencer`?

pkg/controller/cache/redis_test.go
```diff
@@ -184,713 +160,528 @@ func redisResource(rm ...redisResourceModifier) *v1alpha3.Redis {
...
+	type args struct {
+		cr          *v1alpha3.Redis
+		newClientFn func(ctx context.Context, credentials []byte) (redisapi.ClientAPI, error)
+		kube        client.Client
+	}
+	type want struct {
+		err error
+	}
+
+	cases := map[string]struct {
```
negz: I've been playing with a convention if including a `reason string` in each case that is printed when a test fails. The idea is that it makes it more obvious what the test is trying to do to future readers of the test code. So for example the `Successful` test might have `reason: "Connect should not return an error when we can successfully use the Provider and its Secret to produce a new client"`. Do you think it would be worth using this pattern for these tests?

apis/cache/v1alpha3/redis_types.go
jbw976: in my time at MSFT, it was always pronounced as a word, e.g. "skew".  i never heard it pronounced as the letters, e.g. S-K-U.

therefore, ergo, ipso facto, IMO it should be "A SKU" ("a skew")

pkg/controller/cache/redis_test.go
muvaf: I feel like future code reader's effort to understand what the test does by looking at the case's name is less than the writer's effort to write a detailed reason section for each test. One might think writer does it only once so it's fine; but they'd do it for each and every case and reader doesn't go through all of them but only the ones they need, probably the ones that failed. 

Another note: getting people write tests is already hard, let's just not make it any harder.

pkg/controller/cache/redis.go
muvaf: Probably, it was probably either too early in the morning or too late in the night when I unconsciously decided to change that. Corrected.

pkg/controller/azure.go
muvaf: I'm actually leaned towards having packaged on the claim level. Like `pkg/controller/cache` would have `redis.go`, `memcached.go` managed controllers alongside with one `claim.go` file.

BTW, `RedisController{}` name makes more sense. I remember Goland convinced me that its prefix conflicts with a package name but this seems to be gone now.

apis/cache/v1alpha3/redis_types.go
muvaf: Googling it has shown me that the world is divided into two pieces about that 3-letter word's pronunciation. I'm choosing to opt for the one that is more appealing to the ones who have no idea about what SKU is; _AN SKU_.

pkg/controller/cache/redis.go
muvaf: yes, done.

apis/cache/v1alpha3/redis_types.go
muvaf: It's done. Removed the TODO.

pkg/controller/cache/redis_test.go
negz: Hrm, I don't agree with this philosophy. I think it's generally better to optimise for code readers rather than writers.

Typically many more people will read and need to understand code during its lifetime than people will read it. For our tests we follow the Go philosophy of using (apart from `cmp`) just Go to write tests. I think this is the right choice because it means anyone who can understand Go can figure out our tests, but I do think it makes the _intent_ of each test a little harder to understand compared to testing DSLs that read like `Assert(Foo).Returns(NoError)` or similar. I think this has resulted in some cases us trying to encode our expectations in really long and complicated test names.

That said, this need not block this PR. I am happy to continue to try to sell this idea elsewhere. ;)

apis/cache/v1beta1/redis_types.go
```diff
@@ -14,143 +14,181 @@ See the License for the specific language governing permissions and
...
+	// before making any calls to Azure API.
+
 	// ResourceGroupName in which to create this resource.
-	ResourceGroupName string `json:"resourceGroupName"`
+	// +immutable
+	ResourceGroupName string `json:"resourceGroupName,omitempty"`
+
+	// ResourceGroupNameRef to fetch resource group name.
+	// +immutable
+	ResourceGroupNameRef *ResourceGroupNameReferencerForRedis `json:"ResourceGroupNameRef,omitempty"`
```
negz:
```suggestion
	ResourceGroupNameRef *ResourceGroupNameReferencerForRedis `json:"resourceGroupNameRef,omitempty"`
```

apis/cache/v1beta1/redis_types_test.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 07 November 2019 Muvaffak Onus.
```
negz: Are you particularly proud of this specific test file? ;)

(This needs to be changed to "The Crossplane Authors" similar to the other license headers.)

muvaf: No, not particularly :) I guess Goland defaulted to that header automatically when I created the file. Thanks for catching that!

negz:
@hasheddan PTAL. I can't merge until you resolve your request for changes.

PR #62, opened 2019-11-07
Pin code generation tools using Go modules

negz:

Note that this includes CRD generation, which has been subsumed by make generate in order to ensure we use the same controller-tools to generate Go code and CRDs.

### Checklist

I have:


negz:
Oy. I think the AKS unit tests require kubebuilder to be installed, so this will need further fixes before it can be merged.

negz:
@hasheddan looks like this is now working.

PR #64, opened 2019-11-12
Azure SQL v1beta1

### LINKED ISSUE #28: Azure SQL Server v1beta1 Quality
Azure SQL Server should be refactored to use generic managed reconciler and adopt patterns in https://github.com/crossplaneio/crossplane/pull/840

- [ ] MySQL and PostgreSQL to use generic managed reconciler & to have v1beta1 API quality.
- [ ] YAML updates in all docs that refer to these two resources.

Part of https://github.com/crossplaneio/crossplane/issues/863

muvaf:

This PR tries to refactor our Azure MySQL and PostgreSQL resources to v1beta1 quality by applying patterns [here](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md) and filling the missing things like Update.

Fixes #28 

### Checklist

I have:


pkg/controller/database/mysqlserver/managed.go
```diff
@@ -103,102 +103,98 @@ func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (resource.
...
 
 func (e *external) Update(ctx context.Context, mg resource.Managed) (resource.ExternalUpdate, error) {
-	// TODO(negz): Support updating Azure SQL servers. :)
-
-	return resource.ExternalUpdate{}, nil
+	cr, ok := mg.(*v1alpha3.MySQLServer)
+	if !ok {
+		return resource.ExternalUpdate{}, errors.New(errNotMySQLServer)
+	}
+	// TODO(muvaf): If an async update operation is ongoing, state is still ready
```
muvaf: We need to check with the `Future` object whether the operation is still ongoing as received `mysql.Server` object is misleading. But we don't have access to that `Future` object that is created by the `Update` call in the next reconciles. I've tried a few things but not much success so far. Basically, we're dealing two different source of truth for the state and one is really hard to know.

apis/database/v1alpha3/sql_types.go
```diff
@@ -228,89 +228,116 @@ type SQLServerClassList struct {
...
-	runtimev1alpha1.ResourceStatus `json:",inline"`
-
-	// State of this SQLServer.
-	State string `json:"state,omitempty"`
-
-	// A Message containing detail on the state of this SQLServer, if any.
-	Message string `json:"message,omitempty"`
+// SQLServerObservation represents the current state of Azure SQL resource.
+type SQLServerObservation struct {
+	// ID - READ-ONLY; Resource ID
```
negz: Should we strip these `READ-ONLY` notes out? I think this may be confusing in the context of a Crossplane user reading the CRD or API documentation generated from these comments, given that everything in the observation is read only from their perspective.

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+type MySQLServerAPI interface {
+	ServerNameTaken(ctx context.Context, s *azuredbv1alpha3.MySQLServer) (bool, error)
+	GetServer(ctx context.Context, s *azuredbv1alpha3.MySQLServer) (mysql.Server, error)
+	CreateServer(ctx context.Context, s *azuredbv1alpha3.MySQLServer, adminPassword string) error
+	UpdateServer(ctx context.Context, s *azuredbv1alpha3.MySQLServer) error
+	DeleteServer(ctx context.Context, s *azuredbv1alpha3.MySQLServer) error
+}
+
+//---------------------------------------------------------------------------------------------------------------------
+// MySQLServerClient
```
negz: I'd personally be happy to see these "demarcation" comment headers go away - I think they're unique to some of the older Azure codebase. If we think this file is long and complex enough that we need comments like these to make it more navigable we might consider instead breaking it up into multiple files.

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+
+// UpdateServer updates a MySQL Server.
+func (c *MySQLServerClient) UpdateServer(ctx context.Context, cr *azuredbv1alpha3.MySQLServer) error {
+	// TODO(muvaf): password update via Update call is supported by Azure but
+	// we don't support that.
+	s := cr.Spec.ForProvider
+	properties := &mysql.ServerUpdateParametersProperties{
+		Version:        mysql.ServerVersion(s.Version),
+		SslEnforcement: mysql.SslEnforcementEnum(s.SSLEnforcement),
+		//ReplicationRole: s.Spec.ForProvider.ReplicationRole,
```
negz: Should this be commented out?

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+
+// UpdateServer updates a PostgreSQL Server.
+func (c *PostgreSQLServerClient) UpdateServer(ctx context.Context, cr *azuredbv1alpha3.PostgreSQLServer) error {
+	// TODO(muvaf): password update via Update call is supported by Azure but
+	// we don't support that.
+	s := cr.Spec.ForProvider
+	properties := &postgresql.ServerUpdateParametersProperties{
+		Version:        postgresql.ServerVersion(s.Version),
+		SslEnforcement: postgresql.SslEnforcementEnum(s.SSLEnforcement),
+		//ReplicationRole: s.Spec.ForProvider.ReplicationRole,
```
negz: Should this be commented out?

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+	v.Status.State = string(az.VirtualNetworkRuleProperties.State)
+	v.Status.ID = azure.ToString(az.ID)
+	v.Status.Type = azure.ToString(az.Type)
+}
+
+// Helper functions
+// NOTE: These helper functions work for both MySQL and PostreSQL, but we cast everything to the MySQL types because
+// the generated Azure clients for MySQL and PostgreSQL are exactly the same content, just a different package. See:
+// https://github.com/Azure/azure-sdk-for-go/blob/master/services/mysql/mgmt/2017-12-01/mysql/models.go
+// https://github.com/Azure/azure-sdk-for-go/blob/master/services/postgresql/mgmt/2017-12-01/postgresql/models.go
```
negz: Is this comment about casting to the MySQL types still relevant?

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+
+// The name must match the specification of the SKU, so, we don't allow user
+// to specify an arbitrary name.
+
+// ToMySQLSKU returns the name of the MySQL Server SKU, which is tier + family + cores, e.g. B_Gen4_1, GP_Gen5_8.
```
negz: Does this return just the name? The comment reads a little like it's returning a (string?) name, but it seems to be returning a pointer to struct.

pkg/clients/database/sql.go
```diff
@@ -0,0 +1,612 @@
...
+	if in.StorageProfile != nil {
+		p.StorageProfile.BackupRetentionDays = azure.LateInitializeIntPtrFromInt32Ptr(p.StorageProfile.BackupRetentionDays, in.StorageProfile.BackupRetentionDays)
+		p.StorageProfile.GeoRedundantBackup = azure.LateInitializeStringPtrFromVal(p.StorageProfile.GeoRedundantBackup, string(in.StorageProfile.GeoRedundantBackup))
+		p.StorageProfile.StorageAutogrow = azure.LateInitializeStringPtrFromVal(p.StorageProfile.StorageAutogrow, string(in.StorageProfile.StorageAutogrow))
+	}
+}
+
+// IsMySQLUpToDate is used to report whether given mysql.Server is in
+// sync with the SQLServerParameters that user desires.
+// nolint:gocyclo
```
negz: Nitpick: my preference is to put these `nolint` directives right after the opening brace of the function, so they don't leak into the GoDoc.

pkg/controller/database/mysqlserver/managed.go
```diff
@@ -35,7 +36,7 @@ import (
 
 	"github.com/crossplaneio/stack-azure/apis/database/v1alpha3"
 	azurev1alpha3 "github.com/crossplaneio/stack-azure/apis/v1alpha3"
-	azure "github.com/crossplaneio/stack-azure/pkg/clients"
+	azuresql "github.com/crossplaneio/stack-azure/pkg/clients"
```
negz: Is `github.com/crossplaneio/stack-azure/pkg/clients` really just SQL client code now? Should it be in a package like `pkg/clients/sql`?

pkg/controller/database/mysqlserver/managed.go
```diff
@@ -103,102 +105,98 @@ func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (resource.
...
 type external struct {
-	client        azure.MySQLServerAPI
+	kube          client.Client
+	client        database.MySQLServerAPI
 	newPasswordFn func(len int) (password string, err error)
 }
 
 func (e *external) Observe(ctx context.Context, mg resource.Managed) (resource.ExternalObservation, error) {
-	s, ok := mg.(*v1alpha3.MySQLServer)
+	cr, ok := mg.(*v1alpha3.MySQLServer)
```
negz: I notice you use `cr` as the variable name for most managed resources. What does `cr` stand for?

pkg/controller/database/mysqlserver/managed.go
```diff
@@ -103,102 +105,98 @@ func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (resource.
...
-		}
-		return resource.ExternalObservation{ResourceExists: false}, nil
+		return resource.ExternalObservation{ResourceExists: creating}, nil
 	}
 	if err != nil {
 		return resource.ExternalObservation{}, errors.Wrap(err, errGetMySQLServer)
 	}
+	database.LateInitializeMySQL(&cr.Spec.ForProvider, server)
+	if err := e.kube.Update(ctx, cr); err != nil {
+		return resource.ExternalObservation{}, err
```
negz: Should we wrap this error?

pkg/clients/database/sql.go
muvaf: Yes, `ReplicationRole` is not used in `DefaultCreateMode` which is the only mode we currently support. But since `ServerUpdateParametersProperties` struct is used for all modes, it has that field.

pkg/clients/database/sql.go
muvaf: See https://github.com/crossplaneio/stack-azure/pull/64#discussion_r345739108

pkg/clients/database/sql.go
muvaf: It's not. Just changed it.

pkg/controller/database/mysqlserver/managed.go
muvaf: It stands for `Custom Resource`

pkg/controller/database/mysqlserver/managed.go
muvaf: No, it's not. Fixed it.

pkg/controller/database/mysqlserver/managed.go
muvaf: Here is the current update flow:
- We call `Update` when there is a drift. It gets accepted and operation starts. We only check once whether operation is going through, right after the `Update` call.
- Updated parameters are immediately reflected on the received object even though the operation is still ongoing. For example, I've increased the storage from 20GB to 25GB and the first received object already had `25GB`. But in the console UI, you still see `20GB`. Only after the operation is completed, `25GB` is reflected on the UI.

I guess console UI uses a different API because the object returned from `Get` call is different than what UI shows. However, I was suspecting that we'll call `Update` call continuously since we can't really know if an operation is ongoing but that doesn't happen since we compare with what we receive from `Get` and that's immediately updated.

It works but the caveat is that we are not able to tell the user that an operation is ongoing and the values in spec might not be the most up-to-date values. Normally, we claim that you can rely to spec fields if the state is ready. But in this case, value in the spec is misleading as `userVisibleState` is always `Ready` even though an operation is still in progress.

muvaf:
@negz this should be ready for merge!

pkg/clients/database/sql.go
negz:
```suggestion
		// ReplicationRole is not used in DefeaultCreateMode.
		//ReplicationRole: s.Spec.ForProvider.ReplicationRole,
```

pkg/clients/database/sql.go
negz:
```suggestion
		// ReplicationRole is not used in DefaultCreateMode.
		//ReplicationRole: s.Spec.ForProvider.ReplicationRole,
```

PR #75, opened 2019-11-25
Track Azure Async Operations

### LINKED ISSUE #69: Opportunistically track Azure SQL operations

CRUD operations in Azure SQL REST API are asynchronous; they [return a URL](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-async-operations) that may be polled to determine whether the requested operation has completed, and whether it was successful.

The initial implementation of the Azure MySQL and PostgreSQL controllers serialised details about these long running operations and stored them in the managed resource's status, using them to make decisions during the next reconcile. The controller has largely ignored these operations since porting to the managed resource reconciler; they are not persisted, and are consulted only once immediately after they are created. This works, but has some caveats:

* We're never quite sure if an operation is running, because there's no way to discover a running operation after it has begun. https://github.com/crossplaneio/stack-azure/pull/64#discussion_r345763687 is an example of this issue.
* We never quite sure if an operation actually succeeded.

We have a solution that works without storing and polling the running operation; I wonder if we can build on this by opportunistically storing any operation in `.status` and checking it at the next reconcile (similar to the original implementation of this controller). If no operation was recorded we'd fall back to the current behaviour. I believe this would be in compliance with API conventions:

> The status summarizes the current state of the object in the system, and is usually persisted with the object by automated processes but may be generated on the fly. At some cost and perhaps some temporary degradation in behavior, the status could be reconstructed by observation if it were lost.

```go
// Returns an Azure-AsyncOperation URL
op.PollingURL()
```

muvaf:

Azure API is designed to work with async operations that are considered independent of the resource that they are operating on. For example, if you call `Create` you don't get to make a `GET` call for the resource and see its status as `creating`; instead you have to track the progress via the `Future` object that's returned from `Create` call by polling it constantly.

This PR keeps the minimum number of properties needed to construct a `Future` object in the status and polls the status in each reconcile. This way we are able to show the progress about the last operation.

Fixes #69 

### Checklist

I have:


pkg/clients/database/mysql.go
```diff
@@ -336,3 +342,42 @@ func IsMySQLUpToDate(p azuredbv1beta1.SQLServerParameters, in mysql.Server) bool
 	}
 	return true
 }
+
+// TODO(muvaf): FetchAsyncOperation can be used by other managed resources as well.
+
+// FetchAsyncOperation updates the given operation object with the most up-to-date
+// status retrieved from Azure API.
+func FetchAsyncOperation(ctx context.Context, client autorest.Sender, as *azuredbv1beta1.AsyncOperation) error {
```
hasheddan: Could we move this to the general `azure` client pkg and maybe move `AsyncOperation` to `azure` api types pkg so that it can be embedded in other resource types?

pkg/clients/database/mysql.go
```diff
@@ -47,6 +50,10 @@ const (
 	StateDisabled = string(mysql.ServerStateDisabled)
 	StateDropping = string(mysql.ServerStateDropping)
 	StateReady    = string(mysql.ServerStateReady)
+
+	AsyncOperationStatusInProgress = "InProgress"
```
hasheddan: What would you think about this (and potentially the state constants) living in the api pkg as opposed to the client pkg?

pkg/clients/database/mysql_test.go
```diff
@@ -313,3 +321,93 @@ func TestUpdateMySQLVirtualNetworkRuleStatusFromAzure(t *testing.T) {
...
+	}
+	type want struct {
+		op  *v1beta1.AsyncOperation
+		err error
+	}
+	cases := map[string]struct {
+		args args
+		want want
+	}{
+		"Should skip when there is no operation": {},
```
hasheddan: Definitely open to push back here, but would you be opposed to actually passing in `nil` operation here? I know this is what is happening with the zero value, but it feels like it might be a little more readable to me :)

pkg/clients/database/mysql.go
muvaf: This one could live in the `api` package but not the state constants unfortunately because they are taken from `mysql` package of Azure. Including them in `api` package would require consumers of the `api` package to have dependency on Azure.

On the other hand, though it'd require a code change when their values change, it makes sense to just get the value as bare string from the library and put in the `api` package. It's very rare that the state strings would change. They could be useful for the type's consumers as they'd want to see the possible states. I'd like to do that in a separate PR.

pkg/clients/database/mysql_test.go
muvaf: Yeah, I feel like @negz will jump and say _hey! this is already `nil`_ 😄

pkg/clients/database/mysql.go
hasheddan: Separate PR sounds good to me :)

pkg/clients/database/mysql_test.go
hasheddan: We'll defer to @negz on this then :)

pkg/clients/database/mysql.go
muvaf: Nevermind, I just pushed another commit for it :)

pkg/clients/database/mysql.go
```diff
@@ -133,26 +139,11 @@ func (c *MySQLServerClient) CreateServer(ctx context.Context, cr *azuredbv1beta1
...
-	// capturing the error. If there is none, we'll never call Create again anyway.
-	// If the problem happens after a while, we'd fallback to old behavior where
-	// we continuously try to create and fail.
-	_, err = op.DoneWithContext(ctx, c.ServersClient.Client)
-	return err
+	cr.Status.AtProvider.LastOperation = v1alpha3.AsyncOperation{
+		PollingURL: op.PollingURL(),
+		Method:     http.MethodPut,
+	}
+	return azure.FetchAsyncOperation(ctx, c.ServersClient.Client, &cr.Status.AtProvider.LastOperation)
```
negz: Would it make sense to have these client level CRUD abstractions return some summary of of the operation, rather than updating the `*azuredbv1beta1.MySQLServer` they're passed? My thinking is:

* We're currently split between calling `azure.FetchAsyncOperation` inside the client and outside of the client (in `Observe`). It could be nice to be consistent and always update the operation from the `ExternalClient`.
* I _think_ most of our "client" types in Crossplane are read-only; they translate the Crossplane type to the cloud provider type and submit it. This makes me feel a (tiny) bit uncomfortable about these methods updating the type they're passed.

pkg/clients/database/mysql.go
```diff
@@ -77,18 +74,27 @@ type MySQLServerClient struct {
 func NewMySQLServerClient(c *azure.Client) (*MySQLServerClient, error) {
 	mysqlServersClient := mysql.NewServersClient(c.SubscriptionID)
 	mysqlServersClient.Authorizer = c.Authorizer
-	mysqlServersClient.AddToUserAgent(azure.UserAgent)
+	if err := mysqlServersClient.AddToUserAgent(azure.UserAgent); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
muvaf: That's the function:
```
if extension != "" {
	c.UserAgent = fmt.Sprintf("%s %s", c.UserAgent, extension)
	return nil
}
return fmt.Errorf("Extension was empty, User Agent stayed as %s", c.UserAgent)
```

And we put a string constant there meaning it's pretty hard to make it return error in the way we call it. But the function says that it can, so, I just wanted to respect it. I thought they might change the underlying implementation and feel safe since the signature didn't change and we get errors on runtime.

pkg/clients/database/mysql.go
muvaf: > It could be nice to be consistent and always update the operation from the ExternalClient.

Yes, makes sense. Moved the fetch calls to `ExternalClient`. IMO, actual `Create/Update/Delete/Get` calls should also be called in `ExternalClient` with the converted objects from client package but that's probably for another PR.

PR #85, opened 2019-12-30
[Backport to Release v0.4.0] runtime: pin to crossplane-runtime v0.3.0

hasheddan:

Backports changes introduced in #83 

### Checklist

I have:


go.sum
```diff
@@ -81,8 +81,8 @@ github.com/crossplaneio/crossplane v0.5.0-rc.0.20191211203849-05517d46525d h1:Xi
 github.com/crossplaneio/crossplane v0.5.0-rc.0.20191211203849-05517d46525d/go.mod h1:f/e99LDWClmhkvMuvK7UMP1g5b718DGmdnpdrflK0HI=
 github.com/crossplaneio/crossplane-runtime v0.2.4-0.20191211004842-fa83d075257c h1:Xmv6RSdmFf2WeBpzVaqEBVxmGVAAUxA61eMgrdQ+Vsw=
```
jbw976: there are still refs to crossplane-runtime v0.2.4.  Should we do something similar to what we did in https://github.com/crossplaneio/stack-rook/pull/32/files#r361492505, or just treat this as a simple backport only that does not introduce additional changes to the release branch outside of the commit being backported?

hasheddan: Hmm, so it looks like this was not just a problem witht `stack-rook`. I think the most consistent thing would be to pin all of stack master branches to crossplane `v0.6.0` in a another commit / PR then backport it as well. Sorry for the confusion here! These are mostly just cosmetic changes, but I agree that consistency is important. Thanks for noticing :) what do you think about this plan?

jbw976: that sounds reasonable to me.  so let's merge this and then you can follow up with that in master (and then backport).  i like the idea of backports remaining (mostly) cherry-picks from the effort in master, when relevant.

PR #88, opened 2020-01-02
sql:  report the error seen during async operation in status directly

### LINKED ISSUE #84: MySQLServer does not report errors that occur in the later stages of patch operation
### What happened?

I created a `MySQLServer` resource with the following SKU:
```
tier: Basic
capacity: 1
family: Gen5
```
Then I wanted to update it to the following:
```
tier: GeneralPurpose
capacity: 2
family: Gen5
```

I always see `InProgress` in `lastOperation` section of `status` but the polling URL keeps changing. Then I logged into Azure portal and saw a bunch of patch requests with error saying something like `tier cannot be changed from Basic`. But I didn't see that error in `status` of `MySQLServer` resource.

### How can we reproduce it?

Create a `MySQLServer` resource with `Basic` tier and after creation, try to change that value to `GeneralPurpose`.

When I debugged the controller, I saw that the Azure server does not return an error in the first fetch async op call. So, it happens after a while. Then in the next reconcile, it fetches the operation and error comes in. However, before having a chance to update `status`,  `Update` is triggered again since the `status` of the last operation is not `InProgress` anymore. Then we start over. That's why the polling URL keeps changing because each time a new `Update` is triggered but the error is not reported in-between `Update`s. 

Crossplane version: 0.6.0
Azure Stack version: 0.4.0

muvaf:

Report the error seen during async operation in status directly since the update operation is renewed immediately if it hits an error.

Fixes #84 

I have a few test cases I'd like to do, so, it's not ready to merge yet.

### Checklist

I have:

### DIFF
--- a/apis/database/v1beta1/sql_types.go
+++ b/apis/database/v1beta1/sql_types.go
@@ -255,6 +255,9 @@ type SQLServerObservation struct {
 	// LastOperation represents the state of the last operation started by the
 	// controller.
 	LastOperation apisv1alpha3.AsyncOperation `json:"lastOperation,omitempty"`
+
+	// LastError is the last error encountered during async operations.
+	LastError string `json:"lastError,omitempty"`
 }
 
 // A SQLServerStatus represents the observed state of a SQLServer.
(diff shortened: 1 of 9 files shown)


hasheddan:
@muvaf just wanted to give you a heads up that I am going to open a PR for pinning master to crossplane `v0.6.0` so that we can cherry-pick backport to stack-azure `v0.4.0` release branch

PR #101, opened 2020-01-31
Plumb up logging and event recording implementations

negz:

https://github.com/crossplaneio/crossplane-runtime/pull/108

This commit updates all stack-aws controllers to use the logging and event recording functionality added in the above crossplane-runtime PR. Legacy controllers are minimally updated to avoid package scoped loggers.

### Checklist

I have:

### DIFF
--- a/pkg/controller/cache/claim.go
+++ b/pkg/controller/cache/claim.go
@@ -18,14 +18,14 @@ package cache
 
 import (
 	"context"
-	"fmt"
-	"strings"
 
 	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	runtimev1alpha1 "github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1"
+	"github.com/crossplaneio/crossplane-runtime/pkg/event"
+	"github.com/crossplaneio/crossplane-runtime/pkg/logging"
 	"github.com/crossplaneio/crossplane-runtime/pkg/reconciler/claimbinding"
 	"github.com/crossplaneio/crossplane-runtime/pkg/reconciler/claimdefaulting"
 	"github.com/crossplaneio/crossplane-runtime/pkg/reconciler/claimscheduling"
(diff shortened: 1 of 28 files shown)


pkg/controller/cache/claim.go
```diff
@@ -35,18 +35,11 @@ import (
...
-func (c *RedisClaimSchedulingController) SetupWithManager(mgr ctrl.Manager) error {
-	name := strings.ToLower(fmt.Sprintf("scheduler.%s.%s.%s",
-		cachev1alpha1.RedisClusterKind,
-		v1beta1.RedisKind,
-		v1beta1.Group))
+// SetupRedisClaimScheduling adds a controller that reconciles RedisCluster
+// claims that include a class selector but omit their class and resource
+// references by picking a random matching Azure Redis class, if any.
+func SetupRedisClaimScheduling(mgr ctrl.Manager, l logging.Logger) error {
+	name := claimscheduling.ControllerName(v1beta1.RedisKind)
```
hasheddan: I think we made the decision [here](https://github.com/crossplaneio/stack-gcp/pull/152/files) to use the name of the claim type for the controller name.

hasheddan: There are some other instances of this that I did not add a comment to in order to avoid comment overload :)

pkg/controller/storage/account/claim.go
```diff
@@ -59,21 +52,16 @@ func (c *ClaimSchedulingController) SetupWithManager(mgr ctrl.Manager) error {
...
-func (c *ClaimDefaultingController) SetupWithManager(mgr ctrl.Manager) error {
-	name := strings.ToLower(fmt.Sprintf("defaulter.%s.%s.%s",
-		storagev1alpha1.BucketKind,
-		v1alpha3.AccountKind,
-		v1alpha3.Group))
+// SetupClaimDefaulting adds a controller that reconciles Bucket claims that
+// omit their resource ref, class ref, and class selector by choosing a default
+// Azure Account resource class if one exists.
+func SetupClaimDefaulting(mgr ctrl.Manager, l logging.Logger) error {
+	name := claimdefaulting.ControllerName(v1alpha3.AccountKind)
```
hasheddan: container / account is the first place where we will run into conflicts with this naming pattern mentioned above. I think it is likely fine to have both claim controllers named as bucket claim controllers (since they are both reconciling them), what do you think?

negz: Yeah, I'm not too concerned here. As best as I can tell this controller name string is only used in logs, events, etc. I imagine we won't treat accounts as a kind of bucket once we solve our claims and composition challenges.

negz:
@hasheddan Thanks for catching that! I believe all claim controllers should now be using their claim kinds. PTAL.

PR #110, opened 2020-02-14
Update Provider with optional credentialsSecretRef

hasheddan:
Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Introduces optional `credentialsSecretRef` on `Provider` and adds checks for each `Connect` function. Also adds unit tests for all changes.

### Checklist

I have:


pkg/controller/storage/container/container.go
```diff
@@ -160,6 +161,10 @@ func (m *containerSyncdeleterMaker) newSyncdeleter(ctx context.Context, c *v1alp
 		return nil, errors.Wrapf(err, "failed to retrieve storage account reference: %s", c.Spec.AccountReference.Name)
 	}
 
+	if acct.GetWriteConnectionSecretToReference() == nil {
+		return nil, errors.New(errAcctSecretNil)
+	}
+
```
hasheddan: Note: this is actually using the `Account`'s `writeConnectionSecretToRef`, not the `Provider` `credentialsSecretRef`. I noticed that we don't check for `nil` here though so i went ahead and updated.

negz: 😬

PR #130, opened 2020-04-06
Support backup and restore of all Azure managed resources

### LINKED ISSUE #96: AKS: Cluster name collision.
Currently, when provisioning AKS cluster we are relying on following to transform AKS cluster name property to conform to Azure naming restrictions. 
```
func SanitizeClusterName(name string) string {
	if len(name) > maxClusterNameLen {
		name = name[:maxClusterNameLen]
	}

	return strings.TrimSuffix(name, "-")
}
```
https://github.com/crossplaneio/crossplane/blob/master/pkg/clients/azure/aks.go#L197

The side effect of this, two clusters could collide on the same name, for example:
```
my-favorite-aks-cluster-which-is-fast
my-favorite-aks-cluster-which-is-slow
```
both will result in `my-favorite-aks-cluster-which-`

negz:

Fixes https://github.com/crossplane/provider-azure/issues/128
Fixes #96 (kind of - by supporting `crossplane.io/external-name` for AKS)

Ensures all Azure managed resources support the `crossplane.io/external-name` annotation:

* [x] `AKSCluster`
* [x] (Storage) `Container`
* [x] (Storage) `Account`
* [x] `ResourceGroup`
* [x] `VirtualNetwork`
* [x] `Subnet`
* [x] `MySQLServerVirtualNetworkRule`
* [x] `PostgreSQLServerVirtualNetworkRule`

### Checklist

I have:


negz:
I've tested the new AKSCluster controller by applying the updated example resource class and claim YAML, and ensuring an AKS cluster was successfully created. I connected to the cluster via kubectl, then deleted it, confirming its AD application (and thus service account) was also deleted.

pkg/clients/compute/aks.go
```diff
@@ -10,205 +10,319 @@ You may obtain a copy of the License at
...
 const (
 	// AgentPoolProfileName is a format string for the name of the automatically
 	// created cluster agent pool profile
 	AgentPoolProfileName = "agentpool"
 
-	maxClusterNameLen = 31
-)
+	// NetworkContributorRoleID lets the AKS cluster managed networks, but not
+	// access them.
+	NetworkContributorRoleID = "/providers/Microsoft.Authorization/roleDefinitions/4d97b98b-1d4f-4787-a291-c67834d212e7"
```
hasheddan, line 49: nitpick: does the Azure SDK store this a `const` anywhere?

pkg/clients/compute/aks.go
```diff
@@ -10,205 +10,319 @@ You may obtain a copy of the License at
...
 
-// AKSSetupClient is a type that implements all of the AKS setup interface
-type AKSSetupClient struct {
-	AKSClusterAPI
-	azure.ApplicationAPI
-	azure.ServicePrincipalAPI
-	authorization.RoleAssignmentsAPI
-}
+	appCredsValidYears = 5
```
hasheddan, line 51: Is this arbitrary?

pkg/clients/compute/aks_test.go
```diff
@@ -1,44 +0,0 @@
-/*
```
hasheddan, line 1: Looks like we don't have any tests for the AKS client. This is not a regression given the current status, are you just thinking we defer adding them since we anticipate a `v1beta1` refactor soon?

pkg/clients/compute/aks.go
negz: Not sure; this was copied over from the previous implementation. I'd prefer to use a constant, but let's do that when we work on migrating this controller to v1beta1 quality.

pkg/clients/compute/aks.go
negz: Seems to be; was copied over from the previous implementation.

pkg/clients/compute/aks_test.go
negz: I am. I'm surprised total coverage went up with this PR - presumably because there's less AKS specific code now that we're using the managed reconciler.

negz:
Okay, I've now tested everything that I touched to make sure it still works as expected. Specifically I:

* Statically provisioned a storage account, then dynamically provisioned a bucket in it.
* Statically provisioned a resource group, network, and subnet, then dynamically provisioned an ASK cluster in the subnet.
* Statically provisioned a resource group, network, and subnet, dynamically provisioned a PostgreSQL server in it, then setup a virtual network rule to allow traffic from the subnet.
* Statically provisioned a resource group, network, and subnet, dynamically provisioned a MySQL server in it, then setup a virtual network rule to allow traffic from the subnet.

Final step before merging is to ensure that all of this can be backed up and restored as expected.

negz:
```
$ velero backup create akstest --include-resources secrets,postgresqlinstances.database.crossplane.io,mysqlinstances.database.crossplane.io,kubernetesclusters.compute.crossplane.io,buckets.storage.crossplane.io,resourcegroups.azure.crossplane.io,virtualnetworks.network.azure.crossplane.io,subnets.network.azure.crossplane.io,accounts.storage.azure.crossplane.io,containers.storage.azure.crossplane.io,aksclusters.compute.azure.crossplane.io,mysqlservers.database.azure.crossplane.io,mysqlservervirtualnetworkrules.database.azure.crossplane.io,postgresqlservers.database.azure.crossplane.io,postgresqlservervirtualnetworkrules.database.azure.crossplane.io
```

Tested by running the above backup, restoring it, and validating that all the restored resources continued to work as expected.

PR #137, opened 2020-04-15
Migrate to simple reference resolution and selection

negz:

This adds a reference selector for all existing references, and uses the new `managed.APISimpleReferenceResolver` to select and resolve references. See https://github.com/crossplane/crossplane-runtime/pull/151 for more detail.

### Checklist

I have:


pkg/controller/storage/account/account.go
```diff
@@ -117,21 +110,6 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
-	if !resource.IsConditionTrue(b.GetCondition(runtimev1alpha1.TypeReferencesResolved)) {
```
hasheddan, line 120: Were we previously resolving references here even though the storage types do not have any ref fields?

negz: Yeah, exactly. I figured it made more sense to remove this logic than to port it to the new resolver.

negz:
I've tested this by applying `examples/cache`, `examples/compute`, and `examples/database`. All resources successfully resolved their references and created. I also tested that error messages appeared as expected if I attempted to create a managed resource before the resource it depended on existed, or before the resource it depended on had actually set the referenced value.

This PR decreases test coverage by a little over 1%, presumably because each referencer function under `apis/` previously had tests. I'm not planning on adding tests for the new `ResolveReferences` methods, as they're all effectively boilerplate calls to `reference.NewAPIResolver`, which has its own tests.

negz:
> has that behavior been tested as well?

@hasheddan I didn't test 100% of them, but I have validated that the code generally works as expected by using a selector to pick a resource group reference for a MySQLServer.

PR #141, opened 2020-04-26
Don't delete resource groups that are already deleting

negz:

It seems that calling delete on a resource group extends the time required to delete the group; presumably some deletion process is restarted? This results in the resource group never actually being deleted.

### Checklist

I have:


negz:
Just tested this and confirmed that the resource group was deleted after a minute or so.

PR #146, opened 2020-05-11
Define Azure firewall rules

negz:

Azure SQL databases are secure by default, and allow no network traffic when they are first created. Network traffic can be allowed by whitelisting an Azure subnetwork, or by providing a public IP address range. Crossplane already supports the former; this PR adds support for the latter.

I've tested this by applying the updated examples and confirming that the newly create firewall rule appears in the Azure Portal. I haven't (yet) confirmed that I can whitelist a public IP and actually c

### Checklist

I have:


apis/database/v1alpha3/firewallrule.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// ID - Resource ID
+	ID string `json:"id,omitempty"`
+
+	// Type - Resource type.
+	Type string `json:"type,omitempty"`
+}
+
+// A PostgreSQLFirewallRuleSpec defines the desired state of a PostgreSQLFirewallRule.
+type PostgreSQLFirewallRuleSpec struct {
```
muvaf, line 47: `forProvider` 😬

apis/database/v1alpha3/firewallrule.go
```diff
@@ -0,0 +1,153 @@
...
+	// StartIPAddress of the IP range this firewall rule allows.
+	StartIPAddress string `json:"startIpAddress"`
+
+	// EndIPAddress of the IP range this firewall rule allows.
+	EndIPAddress string `json:"endIpAddress"`
+}
+
+// A FirewallRuleStatus represents the observed state of a
+// FirewallRule.
+type FirewallRuleStatus struct {
```
muvaf, line 36: `atProvider` 😬

apis/database/v1alpha3/firewallrule.go
```diff
@@ -0,0 +1,153 @@
...
+// +kubebuilder:object:root=true
+
+// A MySQLServerFirewallRule is a managed resource that represents an
+// Azure MySQL Database virtual network rule.
+// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.bindingPhase"
+// +kubebuilder:printcolumn:name="STATE",type="string",JSONPath=".status.state"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,azure}
+type MySQLServerFirewallRule struct {
```
artursouza, line 137: Is this rule generic or specific per DB?

negz: I think it's specific per DB. The types are identical but the Azure API seems to have different endpoints depending on the type of database you're using, for example:

https://docs.microsoft.com/en-us/rest/api/postgresql/firewallrules/createorupdate uses `PUT https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBforPostgreSQL/servers/{serverName}/firewallRules/{firewallRuleName}?api-version=2017-12-01`

https://docs.microsoft.com/en-us/rest/api/mysql/firewallrules/createorupdate uses `PUT https://management.azure.com/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DBforMySQL/servers/{serverName}/firewallRules/{firewallRuleName}?api-version=2017-12-01`

Note `Microsoft.DBforPostgreSQL` in the first URL vs `Microsoft.DBforMySQL` in the second.

We've historically modelled this as two different types in Crossplane - e.g. `MySQLServer` vs `PostgreSQLServer`. This is a little different from other clouds, where the SQL engine variant is typically a spec field like `spec.engine: mysql`.

negz:
> I've tested this by applying the updated examples and confirming that the newly create firewall rule appears in the Azure Portal.

I haven't (yet) confirmed that I can whitelist my own IP and actually connect to a database. I'll do that once this PR is approved, before merging.

negz:
Confirmed I can connect from my laptop at home when I whitelist its IP:

```
$ docker run -it --rm postgres psql -h "postgres-example.postgres.database.azure.com" -U "myadmin@postgres-example" postgres
Password for user myadmin@postgres-example: 
psql (12.0 (Debian 12.0-2.pgdg100+1), server 9.6.16)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

postgres=> exit;

$ kubectl delete -f me.yaml 
postgresqlserverfirewallrule.database.azure.crossplane.io "negz" deleted

$ docker run -it --rm postgres psql -h "postgres-example.postgres.database.azure.com" -U "myadmin@postgres-example" postgres
psql: error: could not connect to server: FATAL:  no pg_hba.conf entry for host "REDACTED", user "myadmin", database "postgres", SSL on
FATAL:  no pg_hba.conf entry for host "REDACTED", user "myadmin", database "postgres", SSL on
```

PR #149, opened 2020-05-17
Location field of ResourceGroup is required

muvaf:

API returns error if location is not given:
```
'create failed: cannot create ResourceGroup: resources.GroupsClient#CreateOrUpdate:
      Invalid input: autorest/validation: validation failed: parameter=parameters.Location
      constraint=Null value=(*string)(nil) details: value can not be null; required
      parameter'
```

### Checklist

I have:


muvaf:
> looks like the composition in the quick start will need to be update

I don't think so. We just add a validation here to reject the creation without that field. If it's there in the composition, it should work as expected.

PR #161, opened 2020-08-12
Fix already exists error in SQL resources

muvaf:

Azure MySQLServer API allows calling `Create` multiple times and it returns `NotFound` until the server is created successfully. We can't call `Create` multiple times until we see the resource because it'd generate a different password but we don't know whether a creation is in progress either. How we worked around this has been checking the name availability of the MySQL server; if it's not available then it's in creation. But this is not a viable solution since we never return error if in fact a completely different MySQL Server exists with the same name but in a different resource group, meaning we can't take over control but we don't return an error either. What happens is that `Update` fails because it cannot find that SQL Server in that specific `ResourceGroup` like the following:
```
update failed: cannot update PostgreSQLServer: postgresql.ServersClient#Update: Failure sending request: StatusCode=404 -- Original Error: Code="ResourceNotFound" Message="The Resource 'Microsoft.DBforPostgreSQL/servers/sqlserverpostgresql' under resource group 'sqlserverpostgresql-rg' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix"
```

Fixes https://github.com/crossplane/provider-azure/issues/160

### Checklist

I have:

### DIFF
--- a/pkg/clients/database/mysql.go
+++ b/pkg/clients/database/mysql.go
@@ -56,7 +56,6 @@ var (
 
 // MySQLServerAPI represents the API interface for a MySQL Server client
 type MySQLServerAPI interface {
-	ServerNameTaken(ctx context.Context, s *azuredbv1beta1.MySQLServer) (bool, error)
 	GetServer(ctx context.Context, s *azuredbv1beta1.MySQLServer) (mysql.Server, error)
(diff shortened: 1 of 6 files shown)


muvaf:
> Based on your PR description and slack it appears you have already tested this manually.

@hasheddan yes, I tested it manually.

PR #171, opened 2020-09-22
Remove claims and classes

muvaf:

See https://github.com/crossplane/crossplane/issues/1670 for context.

### Checklist

I have:


muvaf:
Updated with master versions and manually tested with examples in the repo.

pkg/controller/storage/account/account.go
```diff
@@ -187,7 +187,7 @@ func newAccountSyncDeleter(ao azurestorage.AccountOperations, kube client.Client
 
 func (asd *accountSyncDeleter) delete(ctx context.Context) (reconcile.Result, error) {
 	asd.acct.Status.SetConditions(runtimev1alpha1.Deleting())
-	if asd.acct.Spec.ReclaimPolicy == runtimev1alpha1.ReclaimDelete {
+	if asd.acct.Spec.DeletionPolicy == runtimev1alpha1.DeletionDelete {
```
negz, line 190: Do we need to handle the case where the DeletionPolicy is not set (and should default to delete, iirc)?

negz (approved):
I left one comment, and I think the integration tests will need to be disabled, but otherwise LGTM. For the integration tests I've taken an approach of removing them completely in a separate commit so that we can easily revert and update when the new packages are ready.

hasheddan (approved):

pkg/controller/storage/account/account.go
muvaf: Now it deleted if the policy is empty.

PR #197, opened 2020-11-07
Crossplane resources generated by k8s-infra-gen

matthchr:
**These are not ready for prime time!**

I am not even sure that it makes sense to merge these to master. 

@kasey suggested I send a PR though based on what we have so far. I believe the idea is to start some discussion as to the shape of resources and what more work we still have to do to get things working. It's up to you if you want to merge this at all, since obviously right now it's not "doing" anything.

This was generated with a private build of [k8s-infra-gen](https://github.com/Azure/k8s-infra) based on my [crossplane-hacking](https://github.com/matthchr/k8s-infra/tree/crossplane-hacking) branch, so it's important to understand that the tool that produced this isn't even merged to master of k8s-infra yet.

I have:

It hasn't! (see above for other reasons why you may not want to merge it).


ErikLundJensen:
@matthchr Could you share your "crossplane-hacking" git repo (if it still exists) ?

matthchr:
This has been merged into master of [k8s-infra-gen](https://github.com/Azure/k8s-infra). None of the generated resources are checked in currently, but you can generate them by running: `go run main.go gen-types azure-crossplane.yaml` from the `hack/generator` folder or by running `task crossplane:generate-types` from the root of the repo.

These resources will need an adapter written so that they satisfy the required Crossplane interface, which we haven't yet done. The adapter will need to look something like [this](https://github.com/Azure/k8s-infra/blob/master/hack/generated/pkg/reconcilers/azure_deployment_reconciler.go).

PR #201, opened 2020-11-23
Feature azurefirewall

escortnotice:

Create a Azure Firewall through crossplane and add NAT and Network Rules to restrict traffic.

Issue:  https://github.com/crossplane/provider-azure/issues/198

I have:

- [ ] Read and followed Crossplane's [contribution process].
- [ ] Run `make reviewable test` to ensure this PR is ready for review.

This code has been tested manually, using the configuration files in the "example" folder which can referred for the same.


apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,azure}
+type AzureFirewall struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   AzureFirewallSpec   `json:"spec"`
+	Status AzureFirewallStatus `json:"status,omitempty"`
+	///Properties SecurityGroupPropertiesFormat   `json:"properties,omitempty"`
```
hasheddan, line 246: Looks like this needs to be removed

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
 }
+
+//Azure Firewall Structs
+// +kubebuilder:object:root=true
+// A AzureFirewall is a managed resource that represents an Azure Firewall
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="STATE",type="string",JSONPath=".status.state"
+// +kubebuilder:printcolumn:name="LOCATION",type="string",JSONPath=".spec.location"
+// +kubebuilder:printcolumn:name="RECLAIM-POLICY",type="string",JSONPath=".spec.reclaimPolicy"
```
hasheddan, line 236: nitpick: `reclaimPolicy` is deprecated so we probably shouldn't show it here

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+	ResourceGroupNameRef *runtimev1alpha1.Reference `json:"resourceGroupNameRef,omitempty"`
+
+	// ResourceGroupNameSelector - Select a reference to the the Azure Firewall
+	// resource group.
+	ResourceGroupNameSelector *runtimev1alpha1.Selector `json:"resourceGroupNameSelector,omitempty"`
+
+	// Location - Resource location.
+	Location string `json:"location"`
+
+	//AzureFirewallPropertiesFormat - Properties of AzureFirewall
```
hasheddan, line 275:
```suggestion
	// AzureFirewallPropertiesFormat - Properties of AzureFirewall
```

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+	AzureFirewallPropertiesFormat `json:"properties,omitempty"`
```
hasheddan, line 276: If this is optional it should have `// +optional`

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// Zones - A list of availability zones denoting where the resource needs to come from.
+	Zones []string `json:"zones,omitempty"`
+
+	// Etag - READ-ONLY; Gets a unique read-only string that changes whenever the resource is updated.
+	Etag string `json:"etag,omitempty"`
+
+	// ID - Resource ID.
+	ID string `json:"id,omitempty"`
```
hasheddan, line 289: `// +optional` plus `*string`. Can you actually set the ID at creation time?

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+
+	// Name - READ-ONLY; Resource name.
+	Name string `json:"name,omitempty"`
+
+	// Type - READ-ONLY; Resource type.
+	Type string `json:"type,omitempty"`
+	// Location - Resource location.
```
hasheddan, line 296:
```suggestion
```

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+	Etag string `json:"etag,omitempty"`
+
+	// ID - Resource ID.
+	ID string `json:"id,omitempty"`
+
+	// Name - READ-ONLY; Resource name.
+	Name string `json:"name,omitempty"`
+
+	// Type - READ-ONLY; Resource type.
+	Type string `json:"type,omitempty"`
```
hasheddan, line 295: These should be in status if read only

pkg/clients/network/azurefirewall.go
```diff
@@ -0,0 +1,247 @@
...
+	return networkmgmt.AzureFirewall{
+		Zones:    azure.ToStringArrayPtr(v.Spec.Zones),
+		Etag:     azure.ToStringPtr(v.Spec.Etag),
+		ID:       azure.ToStringPtr(v.Spec.ID),
+		Name:     azure.ToStringPtr(v.Name),
+		Type:     azure.ToStringPtr(v.Spec.Type),
+		Location: azure.ToStringPtr(v.Spec.Location),
+		Tags:     azure.ToStringPtrMap(v.Spec.Tags),
+		AzureFirewallPropertiesFormat: &networkmgmt.AzureFirewallPropertiesFormat{
+			//ApplicationRuleCollections: nil,
```
hasheddan, line 143: Should this be removed?

pkg/controller/network/AzureFirewall/managed.go
```diff
@@ -0,0 +1,149 @@
...
+	if azureclients.IsNotFound(err) {
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, errGetAzureFirewall)
+	}
+
+	if az.Name != nil {
+		azurefirewall.UpdateAzureFirewallStatusFromAzure(v, az)
+	}
```
hasheddan, line 94: Do any spec fields need to be late initialized also?

apis/network/v1alpha3/types.go
```diff
@@ -225,3 +225,249 @@ type SubnetList struct {
...
+// A AzureFirewall is a managed resource that represents an Azure Firewall
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="STATE",type="string",JSONPath=".status.state"
+// +kubebuilder:printcolumn:name="LOCATION",type="string",JSONPath=".spec.location"
+// +kubebuilder:printcolumn:name="RECLAIM-POLICY",type="string",JSONPath=".spec.reclaimPolicy"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,azure}
+type AzureFirewall struct {
```
hasheddan, line 240: Since this is a new resource it should be introduced at `v1alpha1` version

hasheddan (changes requested):
Thanks for working on this @escortnotice! I have added some review comments below. It looks like the resource schema needs to be cleaned up some and tests need to be added. Also, I think you may have accidentally included an already existing commit in your history. Let me know if you have any questions :)

apis/network/v1alpha3/types.go
ShreyNamdeo: Delete the commented code at line 246

apis/network/v1alpha3/types.go
ShreyNamdeo: Removed reclaim policy declaration code.

apis/network/v1alpha3/types.go
ShreyNamdeo: Added space in comment line 275

apis/network/v1alpha3/types.go
ShreyNamdeo: 'AzureFirewallPropertiesFormat' is not optional.

apis/network/v1alpha3/types.go
ShreyNamdeo: Removed extra comment from the code.

apis/network/v1alpha3/types.go
ShreyNamdeo: Could you please give us a reference code or explain the changes needed as we are not aware.

apis/network/v1alpha3/types.go
ShreyNamdeo: We need to specify these properties while creating firewall , thus we kept them in Spec and we have removed the READ-ONLY signature from comment.

pkg/clients/network/azurefirewall.go
ShreyNamdeo: We have removed this commented code.

pkg/controller/network/AzureFirewall/managed.go
ShreyNamdeo: Can you please explain the statement we did not understand the context.

apis/network/v1alpha3/types.go
ShreyNamdeo: In the existing codebase we can see v1beta1 or v1alpha3 outside the network folder and v1alpha3 inside network folder , so do you suggest us to create v1alpha1 module also , if so we will need some direction or help from you as we are not aware of the process for doing that.

hasheddan: @ShreyNamdeo yes creating a new package would be appropriate here 👍

pkg/controller/network/AzureFirewall/managed.go
hasheddan: Late initialization is the process of setting `spec` fields that are not provided by the user when they create the resource, but are set to default values by Azure. Here is an example of doing this: https://github.com/crossplane/provider-azure/blob/45f75a573ad4cc1727cfcee86d616b4892a11f74/pkg/clients/database/mysql.go#L306

apis/network/v1alpha3/types.go
ulucinar: The [Crossplane managed resource API conventions](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md) doc is a good resource that discusses these topics. The convention is to have pointer types for optional fields, and to use the `// +optional` marker comment for them.

PR #220, opened 2021-02-11
Always supply the Azure API with a SQL server create mode

negz:

https://github.com/crossplane/provider-azure/pull/214 managed to sneak some broken unit tests into master (I think due to some quirk of the detect-noop job). There were some cases in which our tests expected us to send an empty-string CreateMode to the Azure API whereas in practice we were sending `Default`. I believe we always want to send one of the valid create modes (i.e. we should never try to send the empty string).

CC @rbtcollins 

I have:


hasheddan (approved):

rbtcollins:
Looks good to me - sorry about that!

PR #233, opened 2021-03-22
Remove Queue and Table related encryption operations

### LINKED ISSUE #232: Storage Account unsuccessfully tries to disable Table Encryption
### What happened?
Creating a storage account with all encryption services enabled results in a status message unable to Disable Table Encryption and fails to reconcile.

```
storage.AccountsClient#Update: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="FeatureNotSupportedForAccount" Message="Disabling Table Encryption is not supported for the account."
```

This results in continuous failed calls to the Azure API

### How can we reproduce it?
Create a storage account like below - note all encryption services are `true`

```yaml
---
apiVersion: storage.azure.crossplane.io/v1alpha3
kind: Account
metadata:
  name: craigteststorageacc
spec:
  resourceGroupName: craig-test-rg
  storageAccountSpec:
    kind: Storage
    location: westeurope
    sku:
      name: Standard_LRS
      tier: Standard
    tags:
      application: crossplane
    properties:
      encryption:
        keySource: "Microsoft.Storage"
        services:
          blob: true
          file: true
          table: true
          queue: true
  providerRef:
    name: azure-provider
  providerConfigRef:
    name: default
  deletionPolicy: Delete
  writeConnectionSecretToRef:
    namespace: craigtest
    name: storageaccount
```

`kubectl describe` the created account and observe the status message.

`kubectl get <storageaccount> -o yaml` and observe:

```yaml
    properties:
      encryption:
        keySource: Microsoft.Storage
        services:
          blob: true
          file: true
```

Note that table and queue encryption services are removed here. It seems Microsoft may have disabled the ability to set encryption for these services (it is enabled by default).

You should also see regular failing update calls in the activity log for the account:

Screenshot 2021-03-22 at 11 45 31

Crossplane version: 1.1
Provider version: 0.15

craigbryson:
Signed-off-by: craigbryson <37695591+craigbryson@users.noreply.github.com>

Fixes #232 

Removes options for table and queue encryption since these cannot be modified. Accounts now reconcile successfully.

There are some potentially breaking changes to CRDs (removal) not sure how best to handle these?

I have:

- [ X] Read and followed Crossplane's [contribution process].
- [ X] Run `make reviewable test` to ensure this PR is ready for review.

Tested locally provisioning a storage account as per the issue - account now reconciles successfully. No failed API operations are shown in the activity log

### DIFF
--- a/apis/storage/v1alpha3/account.go
+++ b/apis/storage/v1alpha3/account.go
@@ -68,12 +68,6 @@ type EnabledEncryptionServices struct {
 
 	// File - The encryption function of the file storage service.
 	File bool `json:"file,omitempty"`
-
-	// Table - The encryption function of the table storage service.
-	Table bool `json:"table,omitempty"`
-
-	// Queue - The encryption function of the queue storage service.
-	Queue bool `json:"queue,omitempty"`
 }
 
 // newEnabledEncryptionServices from the storage equivalent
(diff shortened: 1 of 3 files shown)


muvaf:
Thanks @craigbryson ! 

From the following comment in SDK, I see that these fields are read-only:
```
	// Table - READ-ONLY; The encryption function of the table storage service.
	Table *EncryptionService `json:"table,omitempty"`
	// Queue - READ-ONLY; The encryption function of the queue storage service.
	Queue *EncryptionService `json:"queue,omitempty"`
```

So, it should be fine to remove them.

I wonder if we can show them in status but it seems like this controller is one of the oldest controllers and maybe the only one that doesn't use the generic managed reconciler, hence due a schema and controller change. So, no need for that for now. Added a comment to https://github.com/crossplane/provider-azure/issues/90

muvaf (approved):

PR #234, opened 2021-03-24
Enum fix in SQL

muvaf:

Weirdly, Azure SDK says it expects `TLS12` format while their enum constants are `TLS1_2`. In a manual test for https://github.com/crossplane/provider-azure/issues/230 , I realized our example is not working due to this. This PR removes the enum and let Azure return error they want in case format is wrong.

I have:

Manually.


go.mod
```diff
@@ -29,5 +29,5 @@ require (
 	k8s.io/apimachinery v0.20.1
 	k8s.io/client-go v0.20.1
 	sigs.k8s.io/controller-runtime v0.8.0
-	sigs.k8s.io/controller-tools v0.3.0
```
hasheddan, line 32: @muvaf was bumping version here intentional?

muvaf: Yes, `make reviewable` made some changes I think not compatible with clean CRD target. So, I just bumped it to make it inline with other providers.

hasheddan: Thanks!

hasheddan (approved):

PR #274, opened 2021-07-10
Remove FieldRequired for EnableVMProtection VNET property

antoninbas:

When setting this property to False in the Crossplane VirtualNetwork CR,
the controller finds itself constantly reconciling actual and desired
state. This leads to a lot of instability when trying to create
resources in that VNET, as there is never any convergence.

Adding a few log messages in the provider, I observed that there was
always a mismatch between the EnableVMProtection property for the
desired object (pointer to a boolean value set to false) and the one
for the actual object obtained from the Azure SDK (which is nil).

So it seems that using FieldRequired for EnableVMProtection may not be
accurate, at least not with the version of the SDK currently used by
Crossplane.

Signed-off-by: Antonin Bas <abas@vmware.com>

I have:

Created Vnets with EnableVMProtection set to false and looked at reconciliation logs.

### DIFF
--- a/pkg/clients/network/network.go
+++ b/pkg/clients/network/network.go
@@ -32,7 +32,7 @@ func NewVirtualNetworkParameters(v *v1alpha3.VirtualNetwork) networkmgmt.Virtual
 		Tags:     azure.ToStringPtrMap(v.Spec.Tags),
 		VirtualNetworkPropertiesFormat: &networkmgmt.VirtualNetworkPropertiesFormat{
 			EnableDdosProtection: azure.ToBoolPtr(v.Spec.VirtualNetworkPropertiesFormat.EnableDDOSProtection, azure.FieldRequired),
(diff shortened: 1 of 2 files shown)


vaspahomov:
EnableVMProtection is a legacy flag, that's why now azure sends nil value.
https://stackoverflow.com/questions/67304812/how-do-i-enable-vm-protection-in-my-virtual-network-on-azure-portal/67306089#67306089

muvaf:
@vaspahomov thanks for the explanation! @antoninbas thanks for the change! have you got to test whether this change fixes the problem you're experiencing?

antoninbas:
@muvaf yes I can confirm that I have tested the change with my Azure account and that it resolves the issue. With this change VNET reconciliation was successful.

I also pushed a new commit as I forgot to update a unit test originally...

muvaf (approved):
Thank you!

PR #283, opened 2021-08-18
Add global poll interval flag

### LINKED ISSUE #282: Add global poll interval flag
I'd like to configure the polling interval of Azure. Similar to [GCP](https://github.com/crossplane/provider-gcp/pull/344) and [AWS](https://github.com/crossplane/provider-aws/pull/741)

Implement the polling interval as in the other providers. See also https://github.com/crossplane/crossplane-runtime/issues/253
Document the polling interval and other flags. See also https://github.com/crossplane/provider-aws/issues/808

hasheddan:

Adds a `--poll-interval` flag to provider, which is passed to all controllers to set their polling interval in the reconciliation happy path.

*Note to reviewers: these changes are almost entirely find and replace.*

xref https://github.com/crossplane/crossplane-runtime/issues/253
Fixes #282 

I have:

See https://github.com/crossplane/provider-aws/pull/741 for testing steps.

### DIFF
--- a/pkg/controller/storage/account/account.go
+++ b/pkg/controller/storage/account/account.go
@@ -49,15 +49,13 @@ const (
 	controllerName = "account.storage.azure.crossplane.io"
 	finalizer      = "finalizer." + controllerName
 
-	reconcileTimeout      = 2 * time.Minute
-	requeueAfterOnSuccess = 1 * time.Minute
-	requeueAfterOnWait    = 30 * time.Second
+	reconcileTimeout   = 2 * time.Minute
+	requeueAfterOnWait = 30 * time.Second
 )
 
 var (
-	resultRequeue    = reconcile.Result{Requeue: true}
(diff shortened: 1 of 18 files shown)


pkg/controller/storage/account/account.go
```diff
@@ -67,17 +65,20 @@ type Reconciler struct {
...
 	managed.Initializer
 
+	poll time.Duration
+
 	log logging.Logger
 }
 
 // Setup adds a controller that reconciles Accounts.
-func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
+func Setup(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, poll time.Duration) error {
```
hasheddan, line 74: The passing of the poll interval through the various nested syncers / updaters in the storage controllers is not super clean, but I opted to go for path of least resistance here since these controllers need to be refactored to use the managed reconciler anyway.

turkenh (approved):

PR #284, opened 2021-08-18
Read PostgreSQL admin password from connection secret

### LINKED ISSUE #230: Wrong password in generated connection-secret for PostgreSQLServer
### What happened?
During the creation of the Azure resource PostgreSQLServer the password inside the generated connection-secret is changed multiple times. And sometimes (not deterministic) the final secret contains a wrong password. It seams to be some kind of timing issue.

I observed the secret changes. They looked like this:

1) secret was created and contains only password data-field
2) 2 seconds after creation the password has changed
3) 1 minute after creation the username, endpoint and port gets generated

If I manually delete the secret it gets recreated by crossplane but the password data-field is missing.

### How can we reproduce it?
Create a Azure resource PostgreSQLServer multiple times and verify if connection with the secret can be established.

Crossplane version: 1.1.0

ulucinar:

This PR proposes a fix for the wrong password in the generated connection secret issue discussed in #230. The proposed behavior is backwards-compatible and in case multiple `Creates` are called and if a connection secret is defined, we make sure that the password initially used to make the Azure PostgreSQLServer create call is reused, effectively preventing multiple erroneous password generations and the issue described in #230. 

Fixes #230 

I have:

I have written a test script that does the following in a loop:
1. Apply the claim defined below to provision a PostgreSQLServer
2. Waits until the PostgreSQLServer MR becomes ready
3. Invokes `az postgres server firewall-rule create` against this newly provisioned instance to allow connections from my workstation
4. Invokes `psql` using the credentials and the connection info stored in the connection secret.
5. Deletes the claim
6. Waits until PostgreSQLServer MR is actually removed

With the proposed fix, issue #230 has *not* been observed in 10 iterations of the above loop. However, running the experiment with the head of the master branch, out of 10 trials, 8 were successful and 2 have failed with the issue reported in #230 (`psql: error: FATAL:  password authentication failed for user "myadmin"`).

Manifests used for the experiments are as follows. These are some stripped down versions of the manifests given [here](https://github.com/crossplane/crossplane/discussions/2192) by @haraldatbmw:
```
---
apiVersion: foo.bar/v1alpha1
kind: PostgreSQLInstance
metadata:
  namespace: default
  name: example
spec:
  parameters:
    location: westeurope
    storageGB: 5
    version: "11"
    virtualNetworkSubnetId: /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Network/virtualNetworks/xxx/subnets/xxx
    dbName: example-db
  compositionSelector:
    matchLabels:
      provider: azure
  writeConnectionSecretToRef:
    name: example-db-connection

---
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: compositepostgresqlinstances.azure.foo.bar
  labels:
    provider: azure
spec:
  writeConnectionSecretsToNamespace: crossplane-system
  compositeTypeRef:
    apiVersion: foo.bar/v1alpha1
    kind: CompositePostgreSQLInstance
  patchSets:
  - name: Metadata
    patches:
    - fromFieldPath: metadata.labels
    - fromFieldPath: metadata.annotations[crossplane.io/external-name]
  resources:
  # resource-group where the db-server lives in
  - name: resourcegroup
    base:
      apiVersion: azure.crossplane.io/v1alpha3
      kind: ResourceGroup
      spec: {}
    patches:
    - type: PatchSet
      patchSetName: Metadata
    - fromFieldPath: "spec.parameters.location"
      toFieldPath: "spec.location"
  # db-server
  - name: postgresqlserver
    base:
      apiVersion: database.azure.crossplane.io/v1beta1
      kind: PostgreSQLServer
      spec:
        forProvider:
          createMode: Default
          administratorLogin: myadmin
          resourceGroupNameSelector:
            matchControllerRef: true
          minimalTlsVersion: TLS1_2
          sslEnforcement: Enabled
          sku:
            tier: GeneralPurpose
            capacity: 2
            family: Gen5  
          storageProfile:
            backupRetentionDays: 35
            geoRedundantBackup: Disabled
            storageAutogrow: Disabled
        writeConnectionSecretToRef:
          namespace: crossplane-system
        providerConfigRef:
          name: default
    patches:
    - type: PatchSet
      patchSetName: Metadata
    - fromFieldPath: "metadata.uid"
      toFieldPath: "spec.writeConnectionSecretToRef.name"
      transforms:
      - type: string
        string:
          fmt: "postgresqlserver-admin-%s"
    - fromFieldPath: "spec.parameters.version"
      toFieldPath: "spec.forProvider.version"
    - fromFieldPath: "spec.parameters.location"
      toFieldPath: "spec.forProvider.location"
    - fromFieldPath: "spec.parameters.storageGB"
      toFieldPath: "spec.forProvider.storageProfile.storageMB"
      transforms:
        - type: math
          math:
            multiply: 1024
  # db-server vnet-rule for subnet where AKS lives in
  - name: vnetrule
    base:
      apiVersion: database.azure.crossplane.io/v1alpha3
      kind: PostgreSQLServerVirtualNetworkRule
      spec:
        providerConfigRef:
          name: default
        resourceGroupNameSelector:
          matchControllerRef: true
        serverNameSelector:
          matchControllerRef: true
    patches:
    - type: PatchSet
      patchSetName: Metadata
    - fromFieldPath: "spec.parameters.virtualNetworkSubnetId"
      toFieldPath: "spec.properties.virtualNetworkSubnetId"

---
apiVersion: apiextensions.crossplane.io/v1
kind: CompositeResourceDefinition
metadata:
  name: compositepostgresqlinstances.foo.bar
spec:
  connectionSecretKeys:
  - username
  - password
  - hostname
  - port
  - database
  group: foo.bar
  names:
    kind: CompositePostgreSQLInstance
    plural: compositepostgresqlinstances
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  version:
                    description: PostgreSQL engine version
                    type: string
                    enum: ["9.5", "9.6", "10", "11"]
                  storageGB:
                    type: integer
                  location:
                    description: Geographic location of this PostgreSQL server.
                    type: string
                  virtualNetworkSubnetId:
                    description: Subnet-Id of your AKS cluster
                    type: string
                  dbName:
                    description: The name of the database
                    type: string
                required:
                - version
                - storageGB
                - location
                - virtualNetworkSubnetId
                - dbName
            required:
            - parameters
```

### DIFF
--- a/pkg/controller/database/postgresqlserver/managed.go
+++ b/pkg/controller/database/postgresqlserver/managed.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/services/postgresql/mgmt/2017-12-01/postgresql"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
@@ -52,6 +54,7 @@ const (
 	errGetPostgreSQLServer    = "cannot get PostgreSQLServer"
 	errDeletePostgreSQLServer = "cannot delete PostgreSQLServer"
 	errFetchLastOperation     = "cannot fetch last operation"
+	errGetConnSecret          = "cannot get connection secret"
 )
 
 // Setup adds a controller that reconciles PostgreSQLInstances.
@@ -146,6 +149,22 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	return o, nil
 }
 
+func (e *external) getPassword(ctx context.Context, cr *v1beta1.PostgreSQLServer) (string, error) {
+	if cr == nil || cr.Spec.WriteConnectionSecretToReference.Name == "" || cr.Spec.WriteConnectionSecretToReference.Namespace == "" {
+		return "", nil
+	}
+
+	s := &v1.Secret{}
+	if err := e.kube.Get(ctx, types.NamespacedName{
+		Namespace: cr.Spec.WriteConnectionSecretToReference.Namespace,
+		Name:      cr.Spec.WriteConnectionSecretToReference.Name,
+	}, s); err != nil {
+		return "", errors.Wrap(err, errGetConnSecret)
+	}
+
+	return string(s.Data[xpv1.ResourceCredentialsSecretPasswordKey]), nil
+}
+
 func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.ExternalCreation, error) {
 	cr, ok := mg.(*v1beta1.PostgreSQLServer)
 	if !ok {
@@ -154,7 +173,13 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 
 	cr.SetConditions(xpv1.Creating())
 
-	pw, err := e.newPasswordFn()
+	pw, err := e.getPassword(ctx, cr)
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	if pw == "" {
+		pw, err = e.newPasswordFn()
+	}
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
 	}
--- a/pkg/controller/database/postgresqlserver/managed.go
+++ b/pkg/controller/database/postgresqlserver/managed.go
@@ -150,7 +150,8 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 }
 
 func (e *external) getPassword(ctx context.Context, cr *v1beta1.PostgreSQLServer) (string, error) {
-	if cr == nil || cr.Spec.WriteConnectionSecretToReference.Name == "" || cr.Spec.WriteConnectionSecretToReference.Namespace == "" {
+	if cr.Spec.WriteConnectionSecretToReference == nil ||
+		cr.Spec.WriteConnectionSecretToReference.Name == "" || cr.Spec.WriteConnectionSecretToReference.Namespace == "" {
 		return "", nil
 	}
 
@@ -179,9 +180,9 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 	}
 	if pw == "" {
 		pw, err = e.newPasswordFn()
-	}
-	if err != nil {
-		return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
+		if err != nil {
+			return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
+		}
 	}
 	if err := e.client.CreateServer(ctx, cr, pw); err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreatePostgreSQLServer)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.14'
+  GO_VERSION: '1.16'
   GOLANGCI_VERSION: 'v1.31'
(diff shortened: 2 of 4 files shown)


ulucinar:
We also need the CI pipeline fixes implemented here:
https://github.com/crossplane/provider-aws/pull/782

pkg/controller/database/postgresqlserver/managed.go
```diff
@@ -146,6 +149,23 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	return o, nil
 }
 
+func (e *external) getPassword(ctx context.Context, cr *v1beta1.PostgreSQLServer) (string, error) {
+	if cr == nil || cr.Spec.WriteConnectionSecretToReference == nil ||
```
muvaf, line 153:
```suggestion
	if cr.Spec.WriteConnectionSecretToReference == nil ||
```

nitpick: we'd need to place nil guard to just too many spaces if we were to check existence of the CR. So, we usually assume it's there after the classic `cr, ok := mg.(*v1beta1.PostgreSQLServer)` statement.

pkg/controller/database/postgresqlserver/managed.go
```diff
@@ -154,7 +174,13 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	pw, err := e.getPassword(ctx, cr)
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	if pw == "" {
+		pw, err = e.newPasswordFn()
+	}
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
 	}
```
muvaf, line 186:
```suggestion
	if pw == "" {
		pw, err = e.newPasswordFn()
		if err != nil {
			return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
		}
	}
	
```

pkg/controller/database/postgresqlserver/managed.go
```diff
@@ -154,7 +174,13 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 
 	cr.SetConditions(xpv1.Creating())
 
-	pw, err := e.newPasswordFn()
+	pw, err := e.getPassword(ctx, cr)
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
```
muvaf, line 180: I thought you said it'd save the password to the connection secret before triggering the creation. Is that not the case anymore?

ulucinar: Nope, it was never the case (also not the case when I ran the above set of experiments). We still leave it to the managed reconciler.

pkg/controller/database/postgresqlserver/managed.go
ulucinar: Done.

pkg/controller/database/postgresqlserver/managed.go
ulucinar: I was worrying that reviewers would not like the nested ifs :)
Done.

pkg/controller/database/postgresqlserver/managed.go
muvaf: Got it. So, we save the password after a successful `CreateServer` call via the managed reconciler and if `CreateServer` is ever called again, we'd be getting the password from connection secret.

muvaf (approved):
LGTM! As we discussed, normally we have secret refs for password input that we use to get the password and if we do want to autogenerate it, so far, we've written the generated password to that input secret and then treated it like user input in the subsequent operations. But since some parts of this would require change in the behavior, it's fine to use connection details but it'd be great if we have an issue opened to add input secret ref field.

PR #293, opened 2021-09-20
Feature/add public access in postgresql

G5Olivieri:

Add support to PublicNetworkAccess field in PostgreSQLServer only.

This implementation is initially for internal use within Stone. For this reason that there is only an implementation for PostgreSQL and not MySQL. The burden of this is that the field is available in MySQL due to the structure of the types and is completely ignored in the MySQLServer CRD.

Fixes https://github.com/crossplane/provider-azure/issues/240

Fixes #

I have:

- Unit tests
- Manually tested

### Disclaimer

Before to test see https://github.com/stone-payments/provider-azure/pull/3 talk about a bug that affect this change.

### DIFF
--- a/pkg/clients/database/postgresql.go
+++ b/pkg/clients/database/postgresql.go
@@ -75,18 +75,19 @@ func (c *PostgreSQLServerClient) GetServer(ctx context.Context, cr *azuredbv1bet
 	return c.ServersClient.Get(ctx, cr.Spec.ForProvider.ResourceGroupName, meta.GetExternalName(cr))
 }
 
-// toMySQLProperties converts the CrossPlane ForProvider object to a PostgreSQL Azure properties object
+// toPGSQLProperties converts the CrossPlane ForProvider object to a PostgreSQL Azure properties object
 func toPGSQLProperties(s v1beta1.SQLServerParameters, adminPassword string) postgresql.BasicServerPropertiesForCreate {
 	createMode := pointerToCreateMode(s.CreateMode)
 	switch createMode {
 	case azuredbv1beta1.CreateModePointInTimeRestore:
 		return &postgresql.ServerPropertiesForRestore{
-			MinimalTLSVersion:  postgresql.MinimalTLSVersionEnum(s.MinimalTLSVersion),
-			Version:            postgresql.ServerVersion(s.Version),
-			SslEnforcement:     postgresql.SslEnforcementEnum(s.SSLEnforcement),
-			CreateMode:         postgresql.CreateModePointInTimeRestore,
-			RestorePointInTime: safeDate(s.RestorePointInTime),
-			SourceServerID:     s.SourceServerID,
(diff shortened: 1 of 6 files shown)


pkg/clients/database/postgresql.go
```diff
@@ -172,9 +176,10 @@ func (c *PostgreSQLServerClient) UpdateServer(ctx context.Context, cr *azuredbv1
 	// we don't support that.
 	s := cr.Spec.ForProvider
 	properties := &postgresql.ServerUpdateParametersProperties{
-		Version:           postgresql.ServerVersion(s.Version),
-		MinimalTLSVersion: postgresql.MinimalTLSVersionEnum(s.MinimalTLSVersion),
-		SslEnforcement:    postgresql.SslEnforcementEnum(s.SSLEnforcement),
+		Version:             postgresql.ServerVersion(s.Version),
+		MinimalTLSVersion:   postgresql.MinimalTLSVersionEnum(s.MinimalTLSVersion),
+		SslEnforcement:      postgresql.SslEnforcementEnum(s.SSLEnforcement),
+		PublicNetworkAccess: postgresql.PublicNetworkAccessEnum(azure.ToString(s.PublicNetworkAccess)),
```
ulucinar, line 182: Did you do some manual tests to ensure that Azure actually supports changing the public access mode of an existing PostgreSQL server?

pkg/clients/database/postgresql_test.go
```diff
@@ -431,3 +431,143 @@ func TestPostgreSQLServerFirewallRuleIsUpToDate(t *testing.T) {
 		})
 	}
 }
+
+func TestIsPostgreSQLUpToDate(t *testing.T) {
```
ulucinar, line 435: Thank you for extending the test coverage. Very much appreciated!

ulucinar (commented):
Thank you very much for the PR @G5Olivieri! Could you please make sure all your commits are signed following the guide [here](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits). We also require them to be signed-off using something similar to `git commit -s` (they already are signed-off btw). 

Regarding the bug you refer to in the description, could you please make sure that we have a bug report in the upstream repo, i.e., crossplane/provider-azure?

Thank you!

pkg/clients/database/postgresql.go
G5Olivieri: Yes, I did manual testing for this case and Azure allows and updates this field in an existing PostgreSQL server

ulucinar (commented):
I think you are using Go v1.17 for your builds. Could you please use Go v1.16 as it's the current version in our CI as specified [here](https://github.com/crossplane/provider-azure/blob/33f728dbaadf9c9448d07ea16119d3137023d34d/.github/workflows/ci.yml#L13).

ulucinar:
LGTM, thank you @G5Olivieri!

ulucinar (approved):

PR #297, opened 2021-09-29
Add Application, ServicePrincipal, Role, Group, RoleAssignment support

vaspahomov:
Add support of: 
* Application
* ServicePrincipal
* RoleAssignment
* Role
* ADGroup

This patch allow:
* Create Applications
* Create ServicePrincipal
* Assign roles on ServicePrincipal
* Assign roles on AD groups

I have:


pkg/controller/rbac/application/managed.go
```diff
@@ -0,0 +1,195 @@
...
+		creds[azure.CredentialsKeyActiveDirectoryGraphResourceID])
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot create service principal token")
+	}
+	if err := token.Refresh(); err != nil {
+		return nil, errors.Wrap(err, "cannot refresh service principal token")
+	}
+	ta := autorest.NewBearerAuthorizer(token)
+	ac := graphrbac.NewApplicationsClient(creds[azure.CredentialsKeyTenantID])
+	ac.Authorizer = ta
```
ulucinar, line 106: Why don't we make use of the `autorest.Authorizer` returned from `azureclients.GetAuthInfo`? Does it not work in this case?:
```suggestion
	creds, auth, err := azureclients.GetAuthInfo(ctx, c.client, mg)
	if err != nil {
		return nil, err
	}
	ac := graphrbac.NewApplicationsClient(creds[azure.CredentialsKeyTenantID])
	ac.Authorizer = auth
```

pkg/controller/rbac/application/managed.go
```diff
@@ -0,0 +1,195 @@
...
+		return managed.ExternalObservation{}, errors.New(errNotApplication)
+	}
+	if meta.GetExternalName(s) == "" {
+		return managed.ExternalObservation{}, nil
+	}
+
+	az, err := e.c.Get(ctx, meta.GetExternalName(s))
+	if err != nil {
+		err = errors.Wrap(err, errGetApplication)
+		err = resource.Ignore(azureclients.IsNotFound, err)
```
ulucinar, line 126: Looks like this will never return true. `azure.IsNotFound` does a type assertion for `autorest.DetailedError` on the error passed as its argument. But we will have an `errors.withStack` (a wrapped error) and never have an `autorest.DetailedError`. I would suggest the usual processing done elsewhere here. As an example please refer to:
https://github.com/crossplane/provider-azure/blob/1c0702110ed45feabe2af15a0478e753d1c34d4e/pkg/controller/network/virtualnetwork/managed.go#L98

pkg/controller/rbac/application/managed.go
```diff
@@ -0,0 +1,195 @@
...
+type external struct {
+	c graphrbacapi.ApplicationsClientAPI
+}
+
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	s, ok := mg.(*v1alpha1.Application)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errNotApplication)
+	}
+	if meta.GetExternalName(s) == "" {
```
ulucinar, line 119: Please see the comment on error checks below.

pkg/controller/rbac/application/managed.go
```diff
@@ -0,0 +1,195 @@
...
+		err = errors.Wrap(err, errGetApplication)
+		err = resource.Ignore(azureclients.IsNotFound, err)
+		return managed.ExternalObservation{}, err
+	}
+
+	s.Status.ApplicationID = azure.ToString(az.AppID)
+	s.SetConditions(xpv1.Available())
+	return managed.ExternalObservation{
+		ResourceExists:   true,
+		ResourceUpToDate: true,
```
ulucinar, line 134: Could you please add a TODO for drift detection here? Or, does this resource not support updates to its parameters?

pkg/controller/rbac/application/managed.go
```diff
@@ -0,0 +1,195 @@
...
+	s, ok := mg.(*v1alpha1.Application)
+	if !ok {
+		return errors.New(errNotApplication)
+	}
+	_, err := e.c.Delete(ctx, meta.GetExternalName(s))
+	return errors.Wrap(resource.IgnoreNotFound(err), errDeleteApplication)
+}
+
+const (
+	appCredsValidYears = 5
```
ulucinar, line 184: Should we make this part of the `spec` (i.e., make it a parameter)?

pkg/controller/rbac/roleassignment/managed.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	"github.com/crossplane/provider-azure/apis/rbac/v1alpha1"
+	azure "github.com/crossplane/provider-azure/pkg/clients"
+	azureclients "github.com/crossplane/provider-azure/pkg/clients"
+)
+
+// Error strings.
+const (
+	errNotRoleAssignment    = "managed resource is not an RoleAssignment"
```
ulucinar, line 48:
```suggestion
	errNotRoleAssignment    = "managed resource is not a RoleAssignment"
```

pkg/controller/rbac/serviceprincipal/managed.go
```diff
@@ -0,0 +1,158 @@
...
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+type connecter struct {
+	client client.Client
+}
+
+func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
+	creds, _, err := azureclients.GetAuthInfo(ctx, c.client, mg)
```
ulucinar, line 83: Please see the comment on Application connector. If the authorizer returned from `GetAuthInfo` is not suitable, please add a comment explaining the situation.

pkg/controller/rbac/serviceprincipal/managed.go
```diff
@@ -0,0 +1,158 @@
...
+	s, ok := mg.(*v1alpha1.ServicePrincipal)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errNotServicePrincipal)
+	}
+	if meta.GetExternalName(s) == "" {
+		return managed.ExternalObservation{}, nil
+	}
+	_, err := e.spc.Get(ctx, meta.GetExternalName(s))
+	if err != nil {
+		return managed.ExternalObservation{}, resource.IgnoreNotFound(errors.Wrap(err, errGetServicePrincipal))
```
ulucinar, line 121: Please see the comment on error handling in Application's `external.Observe`. I think we should be doing the not-found-error check via `azureclients.IsNotFound`.

apis/rbac/v1alpha1/types.go
```diff
@@ -0,0 +1,302 @@
...
+	// +immutable
+	IdentifierUris *[]string `json:"identifierUris,omitempty"`
+}
+
+// An ApplicationSpec defines the desired state of an Application.
+type ApplicationSpec struct {
+	xpv1.ResourceSpec `json:",inline"`
+
+	// ApplicationParameters parameters for an application.
+	ApplicationParameters `json:",inline"`
```
ulucinar, line 53: Per Crossplane [API conventions](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md), this field should be under `spec.forProvider` and not inlined.  For an example, please see:
https://github.com/crossplane/provider-azure/blob/1c0702110ed45feabe2af15a0478e753d1c34d4e/apis/database/v1beta1/configuration_types.go#L103

apis/rbac/v1alpha1/types.go
```diff
@@ -0,0 +1,302 @@
...
+	ApplicationIDRef *xpv1.Reference `json:"applicationIDRef,omitempty"`
+
+	// ApplicationIDSelector - Select a reference to the Application id.
+	// +immutable
+	ApplicationIDSelector *xpv1.Selector `json:"applicationIDSelector,omitempty"`
+
+	// AccountEnabled - whether or not the service principal account is enabled
+	// +optional
+	// +immutable
+	AccountEnabled *bool `json:"accountEnabled,omitempty"`
```
ulucinar, line 109: Could you please introduce a `ServicePrincipalParameters` struct containing these fields? Also per Crossplane [API conventions](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md), this field should be under `spec.forProvider` and not inlined.  For an example, please see:
https://github.com/crossplane/provider-azure/blob/1c0702110ed45feabe2af15a0478e753d1c34d4e/apis/database/v1beta1/configuration_types.go#L103

apis/rbac/v1alpha1/types.go
```diff
@@ -0,0 +1,302 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ServicePrincipal `json:"items"`
+}
+
+// An RoleAssignmentSpec defines the desired state of an RoleAssignment.
+type RoleAssignmentSpec struct {
+	xpv1.ResourceSpec `json:",inline"`
+
+	// PrincipalType - type of principal oneof: ['group', 'service-principal']
```
ulucinar, line 146: Could you please introduce a `RoleAssignmentParameters` struct containing these fields? Also per Crossplane [API conventions](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md), this field should be under `spec.forProvider` and not inlined.  For an example, please see:
https://github.com/crossplane/provider-azure/blob/1c0702110ed45feabe2af15a0478e753d1c34d4e/apis/database/v1beta1/configuration_types.go#L103

ulucinar (changes requested):

apis/rbac/v1alpha1/types.go
```diff
@@ -0,0 +1,302 @@
...
+	xpv1.ResourceStatus `json:",inline"`
+}
+
+// +kubebuilder:object:root=true
+
+// A Role is a resource that represents an Role.
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,azure}
+type Role struct {
```
ulucinar, line 248: My understanding is `Role` is not a managed resource, i.e., we do not have the an `ExternalConnecter` and an `ExternalClient` for `Role`s. Thus, we should not treat it as a managed resource and it should not have a CRD. For the currently defined referencer and selector fields for it, you may just leave the ID field. If there is an actual Azure resource corresponding to roles, then in a separate PR we can introduce the `Role` managed resource that actually manipulates Azure role objects.

apis/rbac/v1alpha1/types.go
```diff
@@ -0,0 +1,302 @@
...
+
+// Role is only CRD and have no reconcile controller
+
+// +kubebuilder:object:root=true
+
+// A ADGroup is a resource that represents an ADGroup.
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,azure}
+type ADGroup struct {
```
ulucinar, line 287: My understanding is `ADGroup` is not a managed resource, i.e., we do not have the an `ExternalConnecter` and an `ExternalClient` for `ADGroup`s. Thus, we should not treat it as a managed resource and it should not have a CRD. For the currently defined referencer and selector fields for it, you may just leave the ID field. If there is an actual Azure resource corresponding to AD groups, then in a separate PR we can introduce the `ADGroup` managed resource that actually manipulates Azure AD group objects.

ulucinar (changes requested):

lovelinuxalot:
Any updates when this PR is getting merged?

ulucinar:
Hi @vaspahomov,
Could you please check the failing tests?

vaspahomov:
Hi @ulucinar,
Removed unused code in tests

ulucinar:
> Hi @ulucinar, Removed unused code in tests

Thank you very much @vaspahomov for bearing with us on this PR. Together with @sergenyalcin, we took another pass over the PR. Among the resolved issues, I did not fully understand the logic here:
https://github.com/crossplane/provider-azure/blob/6c8e8ee885d8d1867992df56902aa1aa49f862c1/pkg/controller/rbac/application/managed.go#L105
or,
https://github.com/crossplane/provider-azure/blob/6c8e8ee885d8d1867992df56902aa1aa49f862c1/pkg/controller/rbac/serviceprincipal/managed.go#L84

Could you please shed some light on these probably with comments?

goober:
Do I understand correctly that this change uses the Active Directory Graph API for communication with Azure?

If that is the case it should be noted that those API:s are deprecated and support will be removed on June 30th, 2022.
https://techcommunity.microsoft.com/t5/azure-active-directory-identity/update-your-applications-to-use-microsoft-authentication-library/ba-p/1257363

mikebollandajw:
we are in need of this

PR #309, opened 2021-11-30
Rename serialized ForProvider as "forProvider"

ulucinar:

I have:

- [X] Run `make reviewable test` to ensure this PR is ready for review.

Manually tested using the example manifest provided in the repo, `examples/network/publicipaddress.yaml`.


apis/network/v1alpha3/types.go
```diff
@@ -230,12 +230,11 @@ type SubnetList struct {
 // A PublicIPAddressSpec defines the desired state of a PublicIPAddress.
 type PublicIPAddressSpec struct {
 	xpv1.ResourceSpec `json:",inline"`
-	// +kubebuilder:validation:Required
-	ForProvider PublicIPAddressFormat `json:"properties"`
+	ForProvider       PublicIPAddressProperties `json:"forProvider"`
```
ulucinar, line 233: Per Crossplane conventions, we had better rename the serialized form as `forProvider` instead of `properties`.

apis/network/v1alpha3/types.go
```diff
@@ -271,14 +270,15 @@ type PublicIPAddressFormat struct {
...
+	Tags map[string]string `json:"tags,omitempty"`
 }
 
 // SKU of PublicIPAddress
 type SKU struct {
 	// Name - Name of sku. Possible values include: ['Standard', 'Basic']
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=Standard;Basic
-	Name string `json:"name,omitempty"`
+	Name string `json:"name"`
```
ulucinar, line 281: `spec.forProvider.sku` is optional. However, if a `spec.forProvider.sku` is specified, I think we need to have its name also specified. Thus removing the `omitempty` tag.

apis/network/v1alpha3/types.go
```diff
@@ -301,6 +302,12 @@ type PublicIPAddressStatus struct {
 	Address string `json:"address"`
 }
 
+// A PublicIPAddressStatus represents the observed state of a SQLServer.
+type PublicIPAddressStatus struct {
+	xpv1.ResourceStatus `json:",inline"`
+	AtProvider          PublicIPAddressObservation `json:"atProvider,omitempty"`
```
ulucinar, line 308: Per Crossplane conventions, let's move the observation under to `status.atProvider`.

apis/network/v1alpha3/types.go
```diff
@@ -254,15 +253,18 @@ type PublicIPAddressFormat struct {
 	// PublicIPAllocationMethod - The public IP address allocation method. Possible values include: 'Static', 'Dynamic'
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=Static;Dynamic
+	// +immutable
 	PublicIPAllocationMethod string `json:"allocationMethod"`
 
-	// PublicIPAllocationMethod - The public IP address version. Possible values include: 'IPV4', 'IPV6'
+	// PublicIPAllocationMethod - The public IP address version. Possible values include: 'IPv4', 'IPv6'
```
ulucinar, line 259: Looks like the SDK constants are defined as `IPv4` and `IPv6`.

apis/network/v1alpha3/types.go
```diff
@@ -254,15 +253,18 @@ type PublicIPAddressFormat struct {
...
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=IPV4;IPV6
+	// +kubebuilder:validation:Enum=IPv4;IPv6
+	// +immutable
 	PublicIPAddressVersion string `json:"version"`
 
 	// Location - Resource location.
 	// +optional
```
muvaf, line 266: Is this really optional?

apis/network/v1alpha3/types.go
```diff
@@ -230,12 +230,11 @@ type SubnetList struct {
...
 type PublicIPAddressSpec struct {
 	xpv1.ResourceSpec `json:",inline"`
-	// +kubebuilder:validation:Required
-	ForProvider PublicIPAddressFormat `json:"properties"`
+	ForProvider       PublicIPAddressProperties `json:"forProvider"`
 }
 
-// PublicIPAddressFormat defines properties of the PublicIPAddress.
-type PublicIPAddressFormat struct {
+// PublicIPAddressProperties defines properties of the PublicIPAddress.
```
muvaf, line 236: FYI, looks like there are [many more](https://docs.microsoft.com/tr-tr/rest/api/virtualnetwork/public-ip-addresses/create-or-update) properties that can be specified for `PublicIPAdress`.

apis/network/v1alpha3/types.go
ulucinar: Turns out not to be. Marked it as required.

apis/network/v1alpha3/types.go
ulucinar: We may consider adding them in later PRs unless there is immediate need, what do you think @muvaf?

muvaf: I personally prefer making it complete while I have the context of the resource, otherwise I'd have to load all that later again, but it can be merged as an alpha resource with less properties than full.

ulucinar: It turns out that we are not sure regarding the immediate requirements around this resource. Hence, I've extended the API surface.

pkg/controller/network/publicipaddress/managed.go
```diff
@@ -98,12 +101,17 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(azureclients.IsNotFound, err), errGetPublicIPAddress)
 	}
 
+	network.LateInitializePublicIPAddress(&s.Spec.ForProvider, &az)
+	if err := e.kube.Update(ctx, s); err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, errUpdateCR)
+	}
+
```
muvaf, line 108: We usually update only if there is a diff and used to do `kube.Update` for late-init in old controllers but the recommended way today is to return `ResourceLateInitialized: true` to make managed reconciler handle it.

pkg/clients/network/network.go
```diff
@@ -132,8 +176,155 @@ func UpdateSubnetStatusFromAzure(v *v1alpha3.Subnet, az networkmgmt.Subnet) {
...
+	}
+
+	if !checkSKU(p.SKU, in.Sku) {
+		return false
+	}
+
+	return checkDNSSettings(p.PublicIPAddressDNSSettings, in.PublicIPAddressPropertiesFormat.DNSSettings)
+}
+
+func checkDNSSettings(d *v1alpha3.PublicIPAddressDNSSettings, in *networkmgmt.PublicIPAddressDNSSettings) bool { // nolint:gocyclo
```
muvaf, line 271:
```suggestion
func checkDNSSettings(d *v1alpha3.PublicIPAddressDNSSettings, in *networkmgmt.PublicIPAddressDNSSettings) bool {
```

pkg/clients/network/network.go
```diff
@@ -132,8 +176,155 @@ func UpdateSubnetStatusFromAzure(v *v1alpha3.Subnet, az networkmgmt.Subnet) {
...
+}
+
+// IsPublicIPAddressUpToDate is used to report whether given network.PublicIPAddress is in
+// sync with the PublicIPAddressProperties that the user desires.
+func IsPublicIPAddressUpToDate(p v1alpha3.PublicIPAddressProperties, in networkmgmt.PublicIPAddress) bool {
+	if !cmp.Equal(p.Tags, azure.ToStringMap(in.Tags), cmpopts.EquateEmpty()) {
+		return false
+	}
+
+	if !checkIPPrefixID(p.PublicIPPrefixID, in.PublicIPPrefix) {
```
muvaf, line 252: `check*` sounds too generic IMO. Do you think we can use something like `is*UpToDate` as a name or something that tells what the function actually checks?

pkg/clients/network/network.go
```diff
@@ -132,8 +176,155 @@ func UpdateSubnetStatusFromAzure(v *v1alpha3.Subnet, az networkmgmt.Subnet) {
 // UpdatePublicIPAddressStatusFromAzure updates the status related to the external
 // Azure public ip address in the PublicIPAddressStatus
 func UpdatePublicIPAddressStatusFromAzure(v *v1alpha3.PublicIPAddress, az networkmgmt.PublicIPAddress) {
```
muvaf, line 178: nitpick: I usually go for a `Generate*Observation` and set the returned value to `cr.Status.AtProvider` in the controller logic so that readers don't have to come here to see what's set.

pkg/controller/network/publicipaddress/managed.go
```diff
@@ -122,7 +130,15 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 }
 
 func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
```
muvaf, line 132: Can you add a test for this function? I believe it'd be almost same as `Create` one.

muvaf (approved):
Thanks @ulucinar ! Looks good to merge after resolving non-nitpick comments.

pkg/controller/network/publicipaddress/managed.go
ulucinar: Thanks @muvaf for the `ResourceLateInitialized: true` suggestion. I've removed the embedded controller-runtime client from our external client implementation and now always set `ObservationResult.ResourceLateInitialized` accordingly. 

We can at a later time revisit `provider-azure`'s late-initialization utility functions so that they report whether fields have been late-initialized. 

While working on this, I had also considered using `cmp.Equal` to detect any changes after late-initialization, however, I did not choose to do so as API server already detects noops and does not update the resource. However, leaving update calls to the managed reconciler is I think a good compromise.

ulucinar: I just realized that if we leave this to the managed reconciler, we are losing any status updates. So, I'm reverting back to making the update call in-site as we already do in the repo:
https://github.com/crossplane/provider-azure/blob/8380dd2d85e0391bd02e23c5c47c13f8962c38a7/pkg/controller/database/postgresqlserver/managed.go#L122

So returning `ResourceLateInitialized: true` makes us lose status updates and I'd like to address proper calculation of this in a repo-wide manner.

PR #310, opened 2021-11-30
Ignore minimalTlsVersion mismatch when the sslEnforcement is disabled

### LINKED ISSUE #307: IsUpToDate of PostgreSqlServer returns false for every reconciliation loop
### What happened?

When I create a PostgreSqlServer resource by using the example manifest, I observed that, provider tries to update the resource for every reconcile loop. Then I debugged the provider and I noticed that `IsPostgreSQLUpToDate` function always returns false. So, the controller calls the Update function for every reconcile loop.

It is minimalTLSVersion field that causes the difference between actual resource in cloud and spec of the custom resource. In example manifest, the value of this field is `TLS12`. However, after creation of resource, you will observe that the value of the field in cloud resource is `TLSEnforcementDisabled`. Therefore, `IsPostgreSQLUpToDate` function noticed that difference between actual and desired state and calls Update function.

### How can we reproduce it?

You should create a PostgreSqlServer resource by using the example manifest. Then, by reviewing the provider logs, you can observe that the Update call is constantly being called.

sergenyalcin:
Signed-off-by: Sergen Yalçın <yalcinsergen97@gmail.com>

Fixes #307 

This PR solves the problem that `IsPostgreSQLUpToDate` function always returns false. When the value of `sslEnforcement` is `Disabled`, the `minimalTlsVersion` mismatch is ignored.

I have:

- A PostgreSqlServer was provisioned by using example manifest (value of `sslEnforcement` is `Disabled`). Then I observed that, the `IsPostgreSQLUpToDate` returned true despite there is a mismatch between the values of `minimalTlsVersion`.

- A PostgreSqlServer was provisioned by using example manifest (value of `sslEnforcement` is `Enabled`, value of `minimalTlsVersion` is `TLS1_2`). Then I observed that, cloud resource created with proper values and `IsPostgreSQLUpToDate` did not return false.

### DIFF
--- a/pkg/clients/database/mysql.go
+++ b/pkg/clients/database/mysql.go
@@ -338,7 +338,7 @@ func IsMySQLUpToDate(p azuredbv1beta1.SQLServerParameters, in mysql.Server) bool
 		return false
 	}
 	switch {
-	case p.MinimalTLSVersion != string(in.MinimalTLSVersion):
+	case p.MinimalTLSVersion != string(in.MinimalTLSVersion) && p.SSLEnforcement != string(mysql.SslEnforcementEnumDisabled):
 		return false
 	case p.SSLEnforcement != string(in.SslEnforcement):
 		return false
--- a/pkg/clients/database/postgresql.go
+++ b/pkg/clients/database/postgresql.go
@@ -331,7 +331,7 @@ func IsPostgreSQLUpToDate(p azuredbv1beta1.SQLServerParameters, in postgresql.Se
 		return false
 	}
 	switch {
-	case p.MinimalTLSVersion != string(in.MinimalTLSVersion):
+	case p.MinimalTLSVersion != string(in.MinimalTLSVersion) && p.SSLEnforcement != string(postgresql.SslEnforcementEnumDisabled):
 		return false
 	case p.SSLEnforcement != string(in.SslEnforcement):
 		return false
--- a/examples/database/postgresqlserver.yaml
+++ b/examples/database/postgresqlserver.yaml
@@ -11,8 +11,8 @@ spec:
     resourceGroupNameRef:
       name: example-rg
     location: West US 2
-    minimalTlsVersion: TLS12
-    sslEnforcement: Disabled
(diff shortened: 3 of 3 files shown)


pkg/clients/database/postgresql.go
```diff
@@ -331,7 +331,7 @@ func IsPostgreSQLUpToDate(p azuredbv1beta1.SQLServerParameters, in postgresql.Se
 		return false
 	}
 	switch {
-	case p.MinimalTLSVersion != string(in.MinimalTLSVersion):
+	case p.MinimalTLSVersion != string(in.MinimalTLSVersion) && p.SSLEnforcement != string(postgresql.SslEnforcementEnumDisabled):
```
muvaf, line 334: This controller is almost same with MySQL one. Would you mind updating that one as well?

sergenyalcin: Done!

muvaf (approved):
Thanks @sergenyalcin !

pkg/clients/database/mysql.go
```diff
@@ -338,7 +338,7 @@ func IsMySQLUpToDate(p azuredbv1beta1.SQLServerParameters, in mysql.Server) bool
 		return false
 	}
 	switch {
-	case p.MinimalTLSVersion != string(in.MinimalTLSVersion):
+	case p.MinimalTLSVersion != string(in.MinimalTLSVersion) && p.SSLEnforcement != string(mysql.SslEnforcementEnumDisabled):
```
ulucinar, line 341: @sergenyalcin do we have a validation on `SSLEnforcement` that forces the enum constants defined in the SDK, e.g., `mysql.SslEnforcementEnumDisabled`?

sergenyalcin: Yes we have a validation for this field:

https://github.com/crossplane/provider-azure/blob/3bbc32a92e59cd0a57eeefc1acaf24fc59f51221/apis/database/v1beta1/sql_types.go#L194

PR #314, opened 2021-12-05
Reuse generated password for AKSCluster's Application

### LINKED ISSUE #298: AKS clusters fail to create - they now need a verified domain of the organization
### What happened?

When creating an AKS cluster, the following events are recorded in the `akscluster` description.

```
  Warning  CannotCreateExternalResource  89s (x12 over 97s)  managed/akscluster.compute.azure.crossplane.io  (combined from similar events): cannot create AKSCluster: graphrbac.ApplicationsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="Unknown" Message="Unknown service error" Details=[{"odata.error":{"code":"Request_BadRequest","date":"2021-10-20T00:26:05","message":{"lang":"en","value":"Values of identifierUris property must use a verified domain of the organization or its subdomain: 'https://xyz.aks.crossplane.io'"},"requestId":"017725b3-d03e-4e26-9a3c-04b67ae2b619","values":[{"item":"PropertyName","value":"identifierUris"},{"item":"PropertyErrorCode","value":"HostNameNotOnVerifiedDomain"},{"item":"HostName","value":"https://xyz.aks.crossplane.io"}]}}]
```

### How can we reproduce it?

The output that follows is from `kubectl get akscluster xyz -o yaml`. It can be used to reproduce the issue.

```yaml
apiVersion: compute.azure.crossplane.io/v1alpha3
kind: AKSCluster
metadata:
  annotations:
    crossplane.io/composition-resource-name: aks
    crossplane.io/external-name: xyz
  creationTimestamp: "2021-10-20T00:24:20Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: xyz-
  generation: 2
  labels:
    crossplane.io/claim-name: ""
    crossplane.io/claim-namespace: ""
    crossplane.io/composite: xyz
  name: xyz
  ownerReferences:
  - apiVersion: devopstoolkitseries.com/v1alpha1
    controller: true
    kind: CompositeCluster
    name: xyz
    uid: 297cca23-106c-43c1-982b-d80b3dc1cea5
  resourceVersion: "58359"
  uid: e9b7fbfb-567f-4ede-8538-04a2eca89dbb
spec:
  dnsNamePrefix: dot
  location: eastus
  nodeCount: 3
  nodeVMSize: Standard_D2_v2
  providerConfigRef:
    name: default
  resourceGroupName: xyz
  version: 1.20.7
status:
  conditions:
  - lastTransitionTime: "2021-10-20T00:24:21Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2021-10-20T00:28:57Z"
    message: 'create failed: cannot create AKSCluster: graphrbac.ApplicationsClient#Create:
      Failure responding to request: StatusCode=400 -- Original Error: autorest/azure:
      Service returned an error. Status=400 Code="Unknown" Message="Unknown service
      error" Details=[{"odata.error":{"code":"Request_BadRequest","date":"2021-10-20T00:30:25","message":{"lang":"en","value":"Values
      of identifierUris property must use a verified domain of the organization or
      its subdomain: ''https://xyz.aks.crossplane.io''"},"requestId":"0a42de07-5487-42be-b951-831e3e57f10b","values":[{"item":"PropertyName","value":"identifierUris"},{"item":"PropertyErrorCode","value":"HostNameNotOnVerifiedDomain"},{"item":"HostName","value":"https://xyz.aks.crossplane.io"}]}}]'
    reason: ReconcileError
    status: "False"
    type: Synced
  endpoint: ""
```

* Cloud provider or hardware configuration

Azure

* Kubernetes version (use `kubectl version`)

```yaml
clientVersion:
  buildDate: "2021-09-15T21:31:32Z"
  compiler: gc
  gitCommit: 8b5a19147530eaac9476b0ab82980b4088bbc1b2
  gitTreeState: clean
  gitVersion: v1.22.2
  goVersion: go1.16.8
  major: "1"
  minor: "22"
  platform: darwin/amd64
serverVersion:
  buildDate: "2021-10-05T19:59:14Z"
  compiler: gc
  gitCommit: 724ef700bab896ff252a75e2be996d5f4ff1b842
  gitTreeState: clean
  gitVersion: v1.21.5+k3s2
  goVersion: go1.16.8
  major: "1"
  minor: "21"
  platform: linux/amd64
```

* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift)

AKS

* OS (e.g. from /etc/os-release)

macOS

* Kernel (e.g. `uname -a`)

```
Darwin Viktors-iMac.local 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64
```

ulucinar:

Fixes #298

This PR proposes a change in the AKSCluster's reconciler where we reuse the graph Application's password credential across reconciliations instead of regenerating a brand new password each time.

With this PR, we also change the scheme of the associated graph Application's App ID URI from `https` to `api`. See [here](https://github.com/crossplane/provider-azure/issues/298#issuecomment-972550233) for a detailed account on this matter.

I have:

- [X] Run `make reviewable test` to ensure this PR is ready for review.

Successfully provisioned an AKSCluster in 5 of the 5 trials.

### DIFF
--- a/pkg/controller/compute/managed.go
+++ b/pkg/controller/compute/managed.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -50,6 +52,7 @@ const (
 	errGetAKSCluster    = "cannot get AKSCluster"
 	errGetKubeConfig    = "cannot get AKSCluster kubeconfig"
 	errDeleteAKSCluster = "cannot delete AKSCluster"
+	errGetConnSecret    = "cannot get connection secret"
 )
 
 // SetupAKSCluster adds a controller that reconciles AKSClusters.
@@ -143,14 +146,42 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.New(errNotAKSCluster)
 	}
 	cr.SetConditions(xpv1.Creating())
-	secret, err := e.newPasswordFn()
+
+	pw, err := e.getPassword(ctx, cr)
 	if err != nil {
-		return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
+		return managed.ExternalCreation{}, err
 	}
-	return managed.ExternalCreation{}, errors.Wrap(e.client.EnsureManagedCluster(ctx, cr, secret), errCreateAKSCluster)
+	if pw == "" {
+		pw, err = e.newPasswordFn()
+		if err != nil {
+			return managed.ExternalCreation{}, errors.Wrap(err, errGenPassword)
+		}
+	}
+	return managed.ExternalCreation{
+		ConnectionDetails: managed.ConnectionDetails{
+			xpv1.ResourceCredentialsSecretPasswordKey: []byte(pw),
+		},
+	}, errors.Wrap(e.client.EnsureManagedCluster(ctx, cr, pw), errCreateAKSCluster)
 }
 
-func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+func (e *external) getPassword(ctx context.Context, cr *v1alpha3.AKSCluster) (string, error) {
+	if cr.Spec.WriteConnectionSecretToReference == nil ||
+		cr.Spec.WriteConnectionSecretToReference.Name == "" || cr.Spec.WriteConnectionSecretToReference.Namespace == "" {
+		return "", nil
+	}
+
+	s := &v1.Secret{}
+	if err := e.kube.Get(ctx, types.NamespacedName{
+		Namespace: cr.Spec.WriteConnectionSecretToReference.Namespace,
+		Name:      cr.Spec.WriteConnectionSecretToReference.Name,
+	}, s); err != nil {
+		return "", errors.Wrap(err, errGetConnSecret)
+	}
+
+	return string(s.Data[xpv1.ResourceCredentialsSecretPasswordKey]), nil
+}
+
+func (e *external) Update(_ context.Context, _ resource.Managed) (managed.ExternalUpdate, error) {
 	// TODO(negz): Support updates.
 	return managed.ExternalUpdate{}, nil
 }
--- a/pkg/controller/compute/managed_test.go
+++ b/pkg/controller/compute/managed_test.go
@@ -24,8 +24,11 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2018-03-31/containerservice"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


pkg/controller/compute/managed.go
```diff
@@ -143,14 +146,42 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.New(errNotAKSCluster)
 	}
 	cr.SetConditions(xpv1.Creating())
-	secret, err := e.newPasswordFn()
+
+	pw, err := e.getPassword(ctx, cr)
```
muvaf, line 150: We need to ask user for sensitive information through [an input `Secret`](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#sensitive-input-fields).

pkg/controller/compute/managed.go
```diff
@@ -143,14 +146,42 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		}
+	}
+	return managed.ExternalCreation{
+		ConnectionDetails: managed.ConnectionDetails{
+			xpv1.ResourceCredentialsSecretPasswordKey: []byte(pw),
+		},
+	}, errors.Wrap(e.client.EnsureManagedCluster(ctx, cr, pw), errCreateAKSCluster)
+}
+
+func (e *external) getPassword(ctx context.Context, cr *v1alpha3.AKSCluster) (string, error) {
```
muvaf, line 167: nitpick: in [RDS instance](https://github.com/crossplane/provider-aws/blob/0fab933/pkg/controller/rds/dbinstance/setup.go#L90), we covered the auto-generate scenario by always asking for a secret but then having the controller populate it with the auto-generated one so that the flow of sensitive information is always the same.

pkg/controller/compute/managed_test.go
```diff
@@ -254,6 +254,11 @@ func TestCreate(t *testing.T) {
 			},
 			want: want{
 				err: errors.Wrap(errBoom, errCreateAKSCluster),
+				ec: managed.ExternalCreation{
+					ConnectionDetails: map[string][]byte{
+						"password": {},
```
muvaf, line 259: How does this end up being empty? Don't we generate the password if it's `""`?

pkg/controller/compute/managed.go
ulucinar: Hi @muvaf,
I don't think it makes sense to make this an input parameter for the `AKSCluster`, though for an `Application` MR things would be different. Here, it's more of an output parameter as we provision an Application as a side-effect to support the AKSCluster.

One discussion around this was to make the Application itself part of the spec and not provision an Application if one is already provided. I have asked for a feature request for this as again, using an existing Application for the cluster is a new scenario that's currently not supported, and not sure if this is really something needed in the context of AKSClusters.

pkg/controller/compute/managed_test.go
ulucinar: The `newPasswordFn` used in this test returns an empty string.

muvaf (approved):
Thanks @ulucinar for digging through this challenging bug! Would you mind opening an issue about splitting `AKSCluster` CRD into more granular CRDs as needed, i.e. separate `Application`, `ServicePrincipal` etc. It should be a blocker before we bump it to `v1beta1`.

pkg/controller/compute/managed.go
muvaf: I think the best way to deal with this is to split them apart to different CRDs and reference one from the other but that shouldn't block this PR.

ulucinar:
Thank you @muvaf. Will open an issue discussing some alternative approaches.

PR #315, opened 2021-12-06
Actually cache the go build cache between CI runs

hasheddan:

The build submodule currently overrides XDG_CACHE_HOME in
order to force the Helm 3 to use the .work/helm directory. This causes Go on
Linux machines to use that directory as the build cache as well. We should
adjust this behavior in the build submodule because it is also causing Linux
users to duplicate their build cache, but for now we just make it easier to
identify its location in CI so that we cache between builds.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Updates CI workflow to cache based on go.cachedir to work around the
fact that we override the GOCACHE in our make context due to the
inclusion of helm.mk.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

I have:

See https://github.com/crossplane/crossplane/pull/2742 for more details -- we are seeing drastic improvements (as would be expected now that we are actually caching) in Crossplane CI. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Find the Go Build Cache
         id: go
-        run: echo "::set-output name=cache::$(go env GOCACHE)"
+        run: echo "::set-output name=cache::$(make go.cachedir)"
 
       - name: Cache the Go Build Cache
         uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


ulucinar (approved):

ulucinar:
Thank you @hasheddan for hunting this issue down.

